PR #58, opened 2022-12-21
Readme edits

MichaelMoriarty67:
Tasks still left:

- [x] Add image link.


jsonsivar:
> What is the status on this PR?

I think it might be stale so closing.  We can reopen later.

PR #65, opened 2022-12-27
fix: refetching queries

csosamorales:
- Set `refetchOnWindowFocus: false` so it won't refetch all the time
- Fix warning on console due missing key on list child

### DIFF
--- a/ui/src/index.tsx
+++ b/ui/src/index.tsx
@@ -7,7 +7,9 @@ import { Provider as ReduxProvider } from 'react-redux';
 import { store } from './store';
 import { QueryClient, QueryClientProvider } from 'react-query';
 
-const queryClient = new QueryClient();
+const queryClient = new QueryClient({
+  defaultOptions: { queries: { refetchOnWindowFocus: false } },
+});
 
 const root = ReactDOM.createRoot(
   document.getElementById('root') as HTMLElement
--- a/ui/src/views/home/list.tsx
+++ b/ui/src/views/home/list.tsx
@@ -14,7 +14,7 @@ export const ListSites = () => {
     <Flex gap={10} mt="40px" flexWrap="wrap" justifyContent="center">
       {new Array(data).fill(0).map((_, index) => {
         const id = data - index - 1;
-        return <SiteCard tokenId={id} />;
+        return <SiteCard key={id} tokenId={id} />;
       })}
     </Flex>
   );


ui/src/views/home/list.tsx
```diff
@@ -14,7 +14,7 @@ export const ListSites = () => {
     <Flex gap={10} mt="40px" flexWrap="wrap" justifyContent="center">
       {new Array(data).fill(0).map((_, index) => {
         const id = data - index - 1;
-        return <SiteCard tokenId={id} />;
+        return <SiteCard key={id} tokenId={id} />;
```
csosamorales, line 17: I think we need another key value to assign since setting the index value to a key is not a good practice. Maybe use uuid for this

csosamorales: But, as it's the tokenId itself maybe we can keep it like this

A question. What's gonna happen if we burn an NFT? Cause right now with this logic we assume that there are no burned NFT's

jsonsivar: You'll know a burned token because the owner will be `address(0)` so we can discard those in the list for now. And for later we can talk about if we want to surface them and if so, how (filters, burned trash bin kinda thing, etc).

ui/src/index.tsx
```diff
@@ -7,7 +7,9 @@ import { Provider as ReduxProvider } from 'react-redux';
 import { store } from './store';
 import { QueryClient, QueryClientProvider } from 'react-query';
 
-const queryClient = new QueryClient();
+const queryClient = new QueryClient({
+  defaultOptions: { queries: { refetchOnWindowFocus: false } },
```
jsonsivar, line 11: Was this causing a problem or we just want to be make less fetches to save on resources?

ui/src/views/home/list.tsx
jsonsivar: If we are able to get the latest token Id from the contract, then would we be able to stop using the index as part of the id calculation? `id` would start at that number and then just be `id -1` on the next round.

ui/src/index.tsx
csosamorales: Yeah, this is why it was executing the query every time the window was focused and I think there's no need to do that

ui/src/views/home/list.tsx
csosamorales: Yeah, the thing is that we need to use map to render the cards, and I think zoruka did this so it renders the cards ordered descending

zoruka (approved):

PR #69, opened 2023-01-05
feat: app mirrors

zoruka:


zoruka:
Not ready for review

contracts/FleekERC721.sol
```diff
@@ -41,8 +46,21 @@ contract FleekERC721 is ERC721, FleekAccessControl {
...
+    struct Mirror {
+        uint256 tokenId;
+        uint256 index;
+        uint256 score;
+        address owner;
+    }
+
+    Counters.Counter private _appIds;
     mapping(uint256 => App) private _apps;
+    mapping(string => Mirror) private _mirrors;
```
jsonsivar, line 63: What string value would this be? some unique slug?

contracts/FleekERC721.sol
```diff
@@ -18,6 +18,10 @@ contract FleekERC721 is ERC721, FleekAccessControl {
     event NewTokenExternalURL(uint256 indexed token, string indexed externalURL, address indexed triggeredBy);
     event NewTokenENS(uint256 indexed token, string indexed ENS, address indexed triggeredBy);
 
+    event NewMirror(string indexed mirrorName, uint256 indexed tokenId, address indexed owner);
+    event NewMirrorScore(string indexed mirrorName, uint256 score, address indexed triggeredBy);
```
jsonsivar, line 22: Should we add tokenId to this event? Same with `RemovedMirror`

contracts/FleekERC721.sol
```diff
@@ -41,8 +46,21 @@ contract FleekERC721 is ERC721, FleekAccessControl {
...
+    /**
+     * The stored data for each mirror.
+     * The mirror "score" is used to determine the best mirror to use.
+     * The mirror "score" equals to 0 means that the mirror is not verified.
+     */
+    struct Mirror {
+        uint256 tokenId;
+        uint256 index;
+        uint256 score;
+        address owner;
```
jsonsivar, line 58: Should we add domain here? Also should we add a type to see if its DNS mirror or Fleek zone type

contracts/FleekERC721.sol
```diff
@@ -267,6 +291,164 @@ contract FleekERC721 is ERC721, FleekAccessControl {
...
+    /**
+     * @dev A view function to gether information about a mirror.
+     * It returns a JSON string representing the mirror information.
+     *
+     * Requirements:
+     *
+     * - the mirror must exist.
+     *
+     */
+    function mirror(string memory mirrorName) public view requireMirror(mirrorName) returns (string memory) {
```
jsonsivar, line 357: Maybe we call this getMirrorURI so its specific

jsonsivar (commented):
Wow what a great start. Is the idea that mirror name is the domain they want to use? I think it should be separate so the domain can be changed keeping all the other properties the same in the future.  Also regarding the score, I think we should also keep two states on the mirror struct called contentVerified and dnsVerified which we can flag on and off. They represent the current state or the last checks.  But the score will be a rep of that over time.  I was just thinking it that way but obviously open to other ways and new ideas.

contracts/FleekERC721.sol
zoruka: My idea was have the unique identifier `Mirror Name` so the token id would not be needed here, but we can add it too.

contracts/FleekERC721.sol
zoruka: The domain would be the mapping key identifier.

contracts/FleekERC721.sol
jsonsivar: Maybe hash(token id + index of mirror)? Just something to start with and we can see if review of MVP flows changes anything.

contracts/FleekERC721.sol
zoruka: So it would be also delivered as base64? Here it is just returning a plain json string.

jsonsivar: Ah gotcha.  Yeah something else but still specific. Seeing `mirror` makes it seem like a constructor for mirror.  But just being picky we can change later.  Maybe something like getMirrorJson.

contracts/FleekERC721.sol
zoruka: The index of the mirror may change when some mirror is removed. Thinking about other choice because this identifier is from where people interact with the mirror, that why it fits well. Imagine the scenario where someone wants to verify a domain, the only way to get to this element directly without having to iterate is when the domain is the key of the mapping.

jsonsivar: Yeah I need to think about this some more.  We also don't know what future use cases and queries are needed so just thinking what's a good flexible structure to go with.

contracts/FleekERC721.sol
```diff
@@ -267,6 +291,164 @@ contract FleekERC721 is ERC721, FleekAccessControl {
...
+     * @dev A view function to check if a mirror is verified.
+     * A mirror is verified if its score is greater than 0.
+     *
+     * Requirements:
+     *
+     * - the mirror must exist.
+     *
+     */
+    function isMirrorVerified(string memory mirrorName) public view requireMirror(mirrorName) returns (bool) {
+        return _mirrors[mirrorName].score > 0;
```
EmperorOrokuSaki, line 382: Let's keep verification separate from the score as that can become confusing. In cases where a mirror is delivering the right content and should be verified but has become a subject to intense censorship, the score because of inaccessibility can drop to 0 and that would falsely identify the mirror as unverified.

contracts/FleekERC721.sol
```diff
@@ -267,6 +291,164 @@ contract FleekERC721 is ERC721, FleekAccessControl {
...
+
+    /**
+     * @dev Increases the score of a mirror registry.
+     *
+     * May emit a {NewMirrorScore} event.
+     *
+     * Requirements:
+     *
+     * - the mirror must exist.
+     * - the sender must have the token controller role.
```
EmperorOrokuSaki, line 393: I thought we were accepting score changes from all addresses?

EmperorOrokuSaki (changes requested):
Great start! Aside from Janison's comments, I added a few as well.

hardhat.config.ts
```diff
@@ -52,11 +52,17 @@ const config: HardhatUserConfig = {
...
+    settings: {
+      optimizer: {
+        enabled: true,
+        runs: 200,
+        details: {
+          yul: true,
+        },
       },
-    ],
+      viaIR: true,
```
EmperorOrokuSaki, line 64: I suggest we set `viaIR` and `yul` to `false` so that it follows the same settings as PR 70. Does the code compile with these settings?

EmperorOrokuSaki (approved):
LGTM

hardhat.config.ts
zoruka: It seems to be working with both on `true`, it is compiling and all tests are passing.

contracts/FleekERC721.sol
```diff
@@ -6,10 +6,13 @@ import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
 import "@openzeppelin/contracts/utils/Counters.sol";
 import "@openzeppelin/contracts/utils/Base64.sol";
 import "./FleekAccessControl.sol";
+import "./util/FleekStrings.sol";
```
jsonsivar, line 9: Should we put these into a `libraries` folder?

contracts/FleekERC721.sol
```diff
@@ -18,6 +21,27 @@ contract FleekERC721 is ERC721, FleekAccessControl {
     event NewTokenExternalURL(uint256 indexed token, string indexed externalURL, address indexed triggeredBy);
     event NewTokenENS(uint256 indexed token, string indexed ENS, address indexed triggeredBy);
 
+    event NewAccessPoint(string indexed apName, uint256 indexed tokenId, address indexed owner);
+    event RemoveAccessPoint(string indexed apName, uint256 indexed tokenId, address indexed owner);
+    event ChangeAccessPointScore(
```
jsonsivar, line 26: Should we move this to another score branch and bring it in then? it's not a big deal just if we want to be minimal or not.

contracts/FleekERC721.sol
```diff
@@ -40,56 +65,36 @@ contract FleekERC721 is ERC721, FleekAccessControl {
...
     /**
      * @dev Initializes the contract by setting a `name` and a `symbol` to the token collection.
      */
     constructor(string memory _name, string memory _symbol) ERC721(_name, _symbol) {}
 
     /**
-     * @dev Checks if msg.sender has the role of tokenOwner for a certain tokenId.
+     * @dev Checks if the AccessPoint exists.
      */
-    modifier requireTokenOwner(uint256 tokenId) {
```
jsonsivar, line 54: To double check, we don't need this for any standard right? This was part of ownable which we are not implementing I believe

contracts/FleekERC721.sol
```diff
@@ -278,6 +267,176 @@ contract FleekERC721 is ERC721, FleekAccessControl {
...
+     *
+     * May emit a {NewAccessPoint} event.
+     *
+     * Requirements:
+     *
+     * - the tokenId must be minted and valid.
+     *
+     * IMPORTANT: The payment is not set yet
+     */
+    function addAccessPoint(uint256 tokenId, string memory apName, string memory source) public payable {
```
jsonsivar, line 283: I'm thinking maybe we remove source? If we have tokenId here then the source really comes from that. The source in the UI is to allow someone to create an AP even if an NFA is not created but that we should probably add a new function for.

jsonsivar: And this can be post MVP and depends on what we design for UI too.

contracts/FleekERC721.sol
```diff
@@ -278,6 +267,176 @@ contract FleekERC721 is ERC721, FleekAccessControl {
...
+        // require(msg.value == 0.1 ether, "You need to pay at least 0.1 ETH"); // TODO: define a minimum price
```
jsonsivar, line 284: Maybe we can make this a conditional.  If a global setting for payments is configured it uses that, otherwise it doesn't.

contracts/FleekERC721.sol
```diff
@@ -278,6 +267,176 @@ contract FleekERC721 is ERC721, FleekAccessControl {
...
+    /**
+     * @dev A view function to gether information about an AccessPoint.
+     * It returns a JSON string representing the AccessPoint information.
+     *
+     * Requirements:
+     *
+     * - the AP must exist.
+     *
+     */
+    function accessPoint(string memory apName) public view requireAP(apName) returns (string memory) {
```
jsonsivar, line 342:
```suggestion
    function getAccessPointString(string memory apName) public view requireAP(apName) returns (string memory) {
```

or something specific

contracts/util/FleekStrings.sol
```diff
@@ -0,0 +1,67 @@
...
+     */
+    function toBase64(string memory str) internal pure returns (string memory) {
+        return Base64.encode(bytes(str));
+    }
+
+    /**
+     * @dev Converts FleekERC721.App to a JSON string.
+     * It requires to receive owner address as a parameter.
+     */
+    function toString(FleekERC721.App storage app, address owner) internal view returns (string memory) {
```
jsonsivar, line 33: Should we copy the tokenId owner to the app struct during minting/transfer? This way this library can just pull from the App input.  Otherwise someone can call this with a random owner and get a URI that's not right.  It's an edge case so your decision in the end @zoruka.  Cuz yes, why would someone do that? So more a question of fool proofing it.

hardhat.config.ts
```diff
@@ -52,16 +52,16 @@ const config: HardhatUserConfig = {
...
   },
   solidity: {
-    version: '0.8.7',
+    version: '0.8.12',
     settings: {
       optimizer: {
         enabled: true,
         runs: 200,
         details: {
-          yul: false,
```
jsonsivar, line 61: What do yul and viaIR flags do? Maybe some comments or we link to the reference for this at top of the file with just one link.

jsonsivar:
Only other comment is we should either add foundry tests or make a card/task to add them after.

hardhat.config.ts
daltoncoder: via-ir is Solidity compilers newer way to compile to bytecode. Basically instead of optimizing the solidity=>opcode it optimizes solidity=>yul=>bytecode, but it opens up alot more optimizations. I have had great results on gas and contract size with this. Here are the docs. Even though i did see when i looked that up, it was still considered experimental 8.12. it was considered stable in 8.13 may want to bump up compiler one more version
https://docs.soliditylang.org/en/v0.8.13/ir-breaking-changes.html?highlight=via-ir

EmperorOrokuSaki:
> Only other comment is we should either add foundry tests or make a card/task to add them after.

I'll create a card and pick it up after the subgraph work today.

contracts/FleekERC721.sol
zoruka: This modifier has not being used since we have implemented our own AccessControl module.

contracts/FleekERC721.sol
zoruka: Source removed

contracts/FleekERC721.sol
zoruka: Maybe we could address this to an issue and create other branch to work on it?

contracts/util/FleekStrings.sol
zoruka: Hmmm I'm thinking about it, since the owner is stored in a struct aside of the app metadata, we would need to also update the metadata every time that mint or transfer is run and enforce that this field is updated always. It seems a good idea for this scenario, but still have the tradeoff of make sure that it is always synced. Maybe I could try build this function in another way. But anyways, since this is a internal function it would be wrong just in case of bad development implementation and seems like users could not trick it.

contracts/FleekERC721.sol
zoruka: As it is a base implementation and can be reworked in the future, would it be worth it to remove from this PR and create another one just carrying the code?

contracts/FleekERC721.sol
zoruka: Here I'm following the structure from openzeppelin repo (https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/utils). But we can change it if you want, what do you think is best?

jsonsivar: I'm okay with this.

contracts/FleekERC721.sol
jsonsivar: Agreed

contracts/util/FleekStrings.sol
jsonsivar: Got it, will create a future card for it.

jsonsivar (approved):
LGTM

PR #71, opened 2023-01-10
refactor: strings library

zoruka:


contracts/FleekERC721.sol
```diff
@@ -157,24 +150,7 @@ contract FleekERC721 is ERC721, FleekAccessControl {
...
-                    '{"trait_type": "ENS", "value":"', app.ENS,'"},',
-                    '{"trait_type": "Commit Hash", "value":"', app.builds[app.currentBuild].commitHash,'"},',
-                    '{"trait_type": "Repository", "value":"', app.builds[app.currentBuild].gitRepository,'"},',
-                    '{"trait_type": "Version", "value":"', app.currentBuild.toString(),'"}',
-                ']',
-            '}'
-        );
-
-        return string(abi.encodePacked(_baseURI(), Base64.encode((dataURI))));
+        return string(abi.encodePacked(_baseURI(), app.toString(owner).toBase64()));
```
daltoncoder, line 153: Everything looks good to me. Maybe if this is the only spot we do app.toString() we could just have the library return this encodePacked already

daltoncoder (approved):

PR #81, opened 2023-01-19
chore: button component

csosamorales:


zoruka (changes requested):

ui/src/components/core/icon/custom/github-icon.tsx
```diff
@@ -0,0 +1,19 @@
...
...am4T1OBHFtebh/FRAt3GPrNRovdqfQFH8fIpAj37OG2TORKPjlAwxDMN5DCu02trziB4nT3Eya0w2SCRcW+wekZ2neKeIBG18y5VTxWt8nyppGCBdz/hcK9Ku+A1Bkn3FlIXK8CA/dTcXfe/sBVBxwXy6S7xloSV9duKLJxKyMwaJwy98G1O9fLB70KnBLnh9+35hTqfssI7uPFjseD5By6wpfgkI8yEai/NAKjxiWp+UHRImVSYOA1cT/6xeyMn58jJ7LjoHTdc8TN9y1ydpYyg+T3iGcM9xyMkS/NPyIw7LaYCHyzOKG8oYh14fwi1mrn5invROazzAeZR8nv+jOHMPu5PjeKOZd5fghr32ysjcGad4Hf5y6moVXMdT4frJnZM0d5dcw98rkG+d158rsNIjZ+t1Y+Mz8igT8SsbhwOvX1+9zFnDh4T5Y/fg6Oj5FZXzYgcfjx5ISRrnGNM0jQ+S+Xfxt3AV3KvD6irjEVYbe8R2zuOxuel3VwLmA35XnydxcuIjfmUTKBnaN3IppUTSx25RDkzBC27qb69CY9JNP7ygQKHMUzw7bTgiwLgx4KW8z8gk+RMatGQMFFCRO4KgJxYdtAIVQmTv0tkHHRj8jDZS2Lvdwbyd8xjmOp9JOdwpazyECUa5AxOBM46/pYgC8N3G6vyHpzn6yHEeuEdMfYuKgl54o8BBL0p/AjOmpl0hfWm2skhNlkCls8EJKqLfQ58UpjKHmPIOlTom/uQZnXLDZVoOmD2dha/BTp33Z2dAmKC5tdaFJcDYFJxtVzInInJhXrxWbNpgvWSq2AszHYVHjUalcQiF4dS67zREkQGIDH6zrmDfJ3i+72+ZJMqNTsE0ZylEfICchusZp2GcYQT/awdkVhZb9BNj1EdNxC4UZixHGWPEdssSmMCsNMb4TgtR+SE534ZBmKizafRk6AQ2iXhkWRvwqTiSmyJFhbBsLiXNVF0uZtYVceZYIyBLEhNusa8h8Ok4SUTBulbWjjc1E9RNQZ6OAnxQlC+KZx7HKVx//3dgTP6jXNVIu0Zbi07XCUBjbpizYFBAekz9lm81itoeiyySOytCGH+L8l51zzyjgZM44Cp4EN9qvI2cRAcAE2HnC4+ctaTgEPqCXn9P4F8maix1kg4r4TRyPGWWCLEhiDLZTxfwEGAIg2ItsKhKpcAAAAAElFTkSuQmCC"
```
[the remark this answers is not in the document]
csosamorales: removed since I'm using react icons for github icon

ui/src/components/core/icon/icon.styles.ts
```diff
@@ -0,0 +1,43 @@
...
+
+      rotate: {
+        true: {
+          transform: 'rotate(-180deg)',
+        },
+      },
+    },
+
+    defaultVariants: {
+      size: 'md',
```
zoruka, line 33: I would leave the font size to `font-size: 1em` instead of defining a default size for it, so it would get the size of the parent so it fits better were it is being display instead you set a size.

ui/src/components/wallet-button/wallet-button.tsx
```diff
@@ -41,15 +41,15 @@ const WalletMenu: React.FC = () => {
           _hover={{ bg: 'custom.gray.100' }}
           bg={'custom.gray.200'}
           onClick={handleCopyAccount}
-          icon={<Icon name="copy" />}
+          // icon={<Icon name="copy" />}
```
zoruka, line 44: Remove all the comments if they will not be used anymore. In it will be kept in the future, leave a TODO comment with why.

ui/src/components/core/button/button.utils.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export type ButtonColor = 'gray' | 'blue';
+export type ButtonVariant = 'solid' | 'outline' | 'ghost' | 'link';
+
+export type GetButtonCompoundVariantOptions = {
+  color?: ButtonColor;
+  variant?: ButtonVariant;
+};
+
+export const getButtonCompoundVariant = ({
```
zoruka, line 11: This seems to be style definition and will not be used in any other file than `button.styled.ts` so maybe it makes more sense to keep in the styling file.

zoruka (changes requested):
Overall points:
- We need to find a fix for importing paths asap. Without it the code is not going to be the intended one. Maybe we need to review how storybook needs to be setup.

- Seems like a lot of code is being added even if is not going to be used just to have it pre-setup. I would recommend to build just what you are going to use because probably a lot of work is going to be wasted in things that will not be used or will be changed. When we are building UI components fundamentally we try to imagine all scenarios that they are going to be used but we hardly ever are accurate on it. Probably some components will require some reworks as the needs for the project are shown, so keep the code simple and with less conditionals that maybe will never be reach seems better.

zoruka (approved):

PR #82, opened 2023-01-20
feat: new handlers and entities for our subgraph to support more queries

EmperorOrokuSaki:


EmperorOrokuSaki:
[sc-266]

subgraph/src/fleek-nfa.ts
```diff
@@ -232,4 +236,26 @@ export function handleTransfer(event: TransferEvent): void {
...
+
+    let holder = Holder.load(owner);
+
+    if (!holder) {
+      // Create a new holder entity
+      holder = new Holder(owner);
+    }
+
+    token.owner = owner;
+    token.minted_by = event.transaction.from;
```
daltoncoder, line 255: event.transaction.from is always gonna be the 0 address because this is in your mint if statement. So i think this field should be set to owner as well. 

A simple addition to this handler would be adding an else condition to this block. So when from is not 0 address it would mean this token was actually transfered. And then you update owner field to always keep the token entity accurate with current owner.

EmperorOrokuSaki: It is possible for someone to mint a token on behalf of another address. So I think there can be a situation in which the person who initiated the transaction would not be the same as the person who the token is minted for.

Agree with the second point.

daltoncoder: Even in that situation event.params.from will equal 0 address(or else we wouldn’t even be in this if block). And event.params.to will always equal the first owner

EmperorOrokuSaki: Yes that's true but this line is not `event.params.from`. I'm getting the from address from the transaction `event.transaction.from`

daltoncoder: My mistake you are correct here. Was mixing up the two

jsonsivar:
Closing in favor of #94

PR #84, opened 2023-01-20
chore: input component

csosamorales:
- input text
- input file

Input file:
Desktop Recording 6


ui/src/components/core/input/input-error-message.ts
```diff
@@ -0,0 +1,16 @@
+import { dripStitches } from '../../../theme/stitches';
+
+const { styled } = dripStitches;
+
+export const StyledErrorMessage = styled('span', {
```
zoruka, line 5: Seems like this component could be included also in the `input.styled.ts` file.

ui/src/components/core/input/input-file.tsx
```diff
@@ -0,0 +1,79 @@
...
+  zIndex: '$docked',
+  my: '$1h',
+
+  '&:hover': {
+    borderColor: '$gray8',
+  },
+});
+
+const validateFileSize = (file: File) => {
+  const fileSize = file.size / 1024; // in KB
```
zoruka, line 26: Here the `1024` could be stored as a config constant. Also would be interesting to have at least a comment explaining why we have the size validation to this size.

ui/src/components/core/input/input-file.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return fileSize <= 10;
+};
```
zoruka, line 28:
```suggestion
const DEFAULT_MAX_FILE_SIZE = 10; // in KB

// The file size must be capped to a size that the contract can handle
const validateFileSize = (file: File, maxSize = DEFAULT_MAX_FILE_SIZE): boolean => {
  return file.size  <= 1024 * maxSize;
};

```

ui/src/components/core/input/input-file.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export const InputFileStyled = () => {
+  const inputFileRef = useRef<HTMLInputElement>(null);
+  const [file, setFile] = useState<File | null>(null); //TODO: do it with redux
```
zoruka, line 32: I think this component would be cool to have props like `value` and `onChange` so it turns out to be controlled like a common input element when we use it on parent components. It seems like we could pass props to the Flex wrapper, so on a parent element we can also control the styling to for example fill a row in a form. It seems like passing the state controlling role to the parent is more correct than handling directly to Redux, because in that way we can for an example have multiple file inputs in a form.

ui/src/components/core/input/input-label.ts
```diff
@@ -0,0 +1,24 @@
+import { dripStitches } from '../../../theme/stitches';
+
+const { styled } = dripStitches;
+
+export const StyledInputLabel = styled('label', {
```
zoruka, line 5: Move this also to `input.styled.ts` so we have a single file for input stylings.

ui/src/components/core/input/input.tsx
```diff
@@ -0,0 +1,43 @@
...
+      {label && (
+        <StyledInputLabel {...ownProps}>
+          {label}
+          {required && ' *'}
+        </StyledInputLabel>
+      )}
+      {type === 'file' ? (
+        <InputFileStyled {...ownProps} />
+      ) : (
+        <StyledInput ref={ref} {...ownProps} type={type} />
```
zoruka, line 35: Looking through the components usage, it seems like we need to add a pattern to component naming with prefix or suffix: `StyledInputLabel` `InputFileStyled`. One idea to keep this more organized is to wrap the components in a object or abstract class. For example, if you wrap all the styled components in an object inside `input.styled.ts` you would be able to locate all of them over an usage like: `StyledInput.Label` and `StyledInput.ErrorMessage`. Also we could use `Styled` only when we are declaring styling elements with stitches, `InputFileStyled` is a component that wraps logic, so it could be called just `InputFile`.

zoruka: Following up in the comment above, when wrapping the components in an object or abstract class:
- Objects are smaller structures in the memory than classes because of the prototype that is created;
- Using classes you are able to use namespaces to declare types, so you are able to define a class and a namespace both called with the same name, which improves the dev experience. Namespaces that have only type/interface declarations aren't included in the bundle, which is good to use only in this scenarios.

zoruka (changes requested):

csosamorales:
Part for story started
[sc-306]

ui/src/components/core/input/input.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+  return (
+    <>
+      <Input size="sm" placeholder="Small" />
+      <Input size="md" placeholder="Medium" required />
+      <Input size="md" placeholder="Medium Invaild" aria-invalid />
+      <Input size="lg" placeholder="Large" />
+      <Input size="sm" placeholder="Small" disabled />
+      <Input size="md" placeholder="Medium" disabled />
+      <Input size="lg" placeholder="Large" disabled />
+      {/* <Input label="Logo" size="md" type="file" /> */}
```
zoruka, line 18: Seems like this should be removed?
```suggestion

```

ui/src/components/core/input/input.ts
```diff
@@ -0,0 +1,66 @@
...
+  defaultVariants: {
+    size: 'md',
+  },
+};
+
+export const Input = styled('input', styles);
+
+export const Textarea = styled('textarea', styles);
+
+export const File = StyledInputFile;
```
zoruka, line 66: This `File` component name seems weird. Maybe the component could be called also as `SVGFileInput` or `LogoFileInput` due to the limitations in the component itself.

ui/src/components/form/form.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+        <Form.Error>Input error</Form.Error>
+      </Form.Field>
+      <Form.Field>
+        <Form.Label>Label</Form.Label>
+        <Form.Textarea placeholder="Textarea" />
+        <Form.Error>Textarea error</Form.Error>
+      </Form.Field>
+      <Form.Field css={{ width: '$24' }}>
+        <Form.Label>Label</Form.Label>
+        <Form.File value={file} onChange={(file) => setFile(file)} />
```
zoruka, line 25: This is broken in storybook:
image

zoruka (changes requested):
Other than the other comments:
- There is an empty file on `ui/src/components/form/index.ts`;
- It seems like storybook is not being used as it could be take a look on https://storybook.js.org/docs/react/essentials/controls.

zoruka (approved):

ui/src/components/core/input/input-file.tsx
```diff
@@ -0,0 +1,94 @@
+import { Flex } from '../../layout';
+import { dripStitches } from '../../../theme';
+import { forwardRef, useRef, useState } from 'react';
+import { Icon } from '../icon';
+import { Form } from '../../../components/form/form';
+
+const { styled } = dripStitches;
+
+const BorderInput = styled('div', {
```
jsonsivar, line 9: Shouldn't these come from another file?

jsonsivar (approved):
LGTM but just one comment in case we want to separate the style into another file. But if that is the best practice it's cool.

ui/src/components/core/input/input-file.tsx
zoruka: Not necessarily since it is being used just inside of this component.

PR #87, opened 2023-01-26
fix: fix button stories

csosamorales:
- fix button story
- rename file to follow the same naming across all the project


ui/src/components/core/button/button.stories.tsx
```diff
@@ -24,14 +24,14 @@ export const Default = () => (
     <Button
       colorScheme="blue"
       variant="outline"
-      css={{ py: '$1', borderRadius: '$md' }}
+      css={{ py: '$1', height: '$5', borderRadius: '$md' }}
```
zoruka, line 27: Seems like change the css in the stories will make the component look different on storybook than in the project usage, is that correct?

ui/src/components/core/input/input.ts
```diff
@@ -0,0 +1,109 @@
+// import { Flex } from '../../layout';
```
zoruka, line 1: It seems like there is changes from other PRs in this one.

ui/src/components/layout/flex.styled.tsx
```diff
@@ -0,0 +1,10 @@
+import { dripStitches } from '../../theme/stitches'; //TODO replace with absolute path
```
zoruka, line 1: There are files with suffix `.styled.tsx` and `.styles.tsx` lets keep using a single one for all of them.

ui/src/components/core/button/button.stories.tsx
csosamorales: Yeap, cause this is a md button, but with some specific changes, like [this one](https://www.figma.com/file/98aKyXDUKu83UoFwGy5BGV/NFA-App?node-id=2%3A10577&t=6Sea988a3GTnRqIb-0), so I did this on the storybook to check how it was being renderer, but then on the view the button would look like this (I mean, the code)

ui/src/components/layout/flex.styled.tsx
csosamorales: agree! I would do it

ui/src/components/core/input/input.ts
csosamorales: 🤦🏼‍♀️

jsonsivar (commented):
Just following up on this from our standup ... let's pause any storybook work and do that later.

For the purposes of this PR, its up to you @zoruka @csosamorales if you want to quickly wrap it up and merge it without spending much time on it or close it as-is.

zoruka (approved):

PR #90, opened 2023-01-26
chore: dropdown component

csosamorales:
chore_ input component by csosamorales · Pull Request  84 · fleekxyz_non-fungible-apps


ui/src/components/core/combobox/combobox.tsx
```diff
@@ -0,0 +1,126 @@
...
+  icon?: IconName;
+};
+
+export type ComboboxProps = {
+  items: ComboboxItem[];
+  selectedValue: ComboboxItem | undefined;
+  onChange(option: ComboboxItem): void;
+};
+
+export const Combobox: React.FC<ComboboxProps> = ({
```
zoruka, line 18: It would be cool to split this component in smaller ones, even if it is in the same file, just to make it simpler to understand and follow the code. For example, here we can have a component to hold the input and the search icon and also a component for options. As bigger the render get, harder is to follow and find/fix bugs.

ui/src/components/core/combobox/combobox.tsx
```diff
@@ -0,0 +1,126 @@
...
+                  open
+                    ? 'border-b-0 rounded-t-xl bg-black border-slate6'
+                    : 'rounded-xl'
+                }`}
+                displayValue={(selectedValue: ComboboxItem) =>
+                  selectedValue.label
+                }
+                onChange={(event) => setQuery(event.target.value)}
+                onClick={() => buttonRef.current?.click()}
+              />
```
zoruka, line 69: It would be cool to remove the handlers from inside the render to keep the code cleaner.

ui/src/components/core/combobox/dropdown.tsx
```diff
@@ -0,0 +1,81 @@
...
+  label: string;
+};
+
+export type DropdownProps = {
+  items: DropdownItem[];
+  selectedValue: DropdownItem | undefined;
+  onChange(option: DropdownItem): void;
+};
+
+export const Dropdown: React.FC<DropdownProps> = ({
```
zoruka, line 17: The same comments from previous component, we could split this in smaller components and move the handlers.

zoruka (approved):
The code seems correct, but still have added some comments about cleaning and organizing the code a bit.

zoruka:
An extra comment, as we are not using storybook anymore, it would be cool to keep examples of the component usage while the project still under development.

PR #91, opened 2023-01-27
feature: stepper component

csosamorales:
Considerations:
- I added the mint stepper as an example, it will change while we start to do de integrations (connect to wallet, connect to github, mint functionality)


zoruka (approved):

ui/src/components/layout/stepper/stepper.stories.tsx
```diff
@@ -0,0 +1,48 @@
+import { Button } from '../../core';
```
jsonsivar, line 1: Is this file needed even if we don't want to do any storybook work for now?

jsonsivar (approved):
LGTM with just the question around the .stories file.  Also maybe we create issues or shortcut cards for the TODOs? I can take care of that.

ui/src/components/layout/stepper/stepper.stories.tsx
csosamorales: It's not mandatory, but it helps me to build just the component cause first I just did the stepper part, without all the related card and the tiltes and button, so it's helpful for the component building (specially the styling part)

PR #99, opened 2023-02-01
feat: add controller entity and handlers.

EmperorOrokuSaki:


jsonsivar:
[sc-307]

subgraph/schema.graphql
```diff
@@ -131,13 +131,20 @@ type Transfer @entity(immutable: true) {
...
-  id: Bytes! #transaction hash
+  id: Bytes! # Token ID
+  mint_transaction_hash: Bytes!
   tokenId: BigInt! # uint256
   owner: Holder!
   minted_by: Bytes!
+  controllers: [Controller!]
 }
 
 type Holder @entity {
```
jsonsivar, line 142: Should we change this to Owner to match the contract?

subgraph/src/fleek-nfa.ts
```diff
@@ -20,12 +20,15 @@ import {
   ApprovalForAll,
   CollectionRoleGranted,
   CollectionRoleRevoked,
+  Controller,
+  Holder,
```
jsonsivar, line 24: Should we call this owner instead? Or token owner?

subgraph/src/fleek-nfa.ts
```diff
@@ -203,6 +206,41 @@ export function handleTokenRoleGranted(event: TokenRoleGrantedEvent): void {
...
+      log.debug('CONTROLLER IS GOING TO BE CREATED HERE.', []);
+      controller = new Controller(event.params.toAddress);
+    }
+
+    if (token !== null) {
+      let token_controllers = token.controllers;
+      if (!token_controllers) {
+        token_controllers = [];
+      }
+      token_controllers.push(event.params.toAddress);
```
jsonsivar, line 231: Don't we need to push a Controller object here instead of the raw address?

subgraph/src/fleek-nfa.ts
```diff
@@ -203,6 +206,41 @@ export function handleTokenRoleGranted(event: TokenRoleGrantedEvent): void {
...
+      token.controllers = token_controllers;
+    } else {
+      log.error(
+        'Controller access granted for tokenId {}. THE TOKEN DOES NOT EXIST. FAILED TO UPDATE THE TOKEN ENTITY.',
```
jsonsivar, line 235: Can we change the first sentence to `Handling controller access granted event for tokenId {}`? The way it is now makes it seem like it is this call that did the granting but it's really processing the event for it.

subgraph/src/fleek-nfa.ts
```diff
@@ -219,6 +257,43 @@ export function handleTokenRoleRevoked(event: TokenRoleRevokedEvent): void {
...
+      if (index > -1) {
+        token_controllers.splice(index, 1);
+      }
+
+      // assign the new controllers list
+      token.controllers = token_controllers;
+    } else {
+      // the token does not exist
+      log.error(
+        'Controller access removed from tokenId {}. THE TOKEN DOES NOT EXIST. FAILED TO UPDATE THE TOKEN ENTITY.',
```
jsonsivar, line 288: Same as above note here.

subgraph/schema.graphql
EmperorOrokuSaki: controllers are different from owners. Owner refers to the ERC721 owner of the NFA who can burn and transfer the asset. Controllers can only update information.

subgraph/src/fleek-nfa.ts
EmperorOrokuSaki: If you are referring to Holder -> Owner, yes. That's the change in the other PR.
But controllers are different.

subgraph/src/fleek-nfa.ts
EmperorOrokuSaki: No. Just the address works.

subgraph/src/fleek-nfa.ts
EmperorOrokuSaki: Of course. Will push a commit for this.

subgraph/schema.graphql
jsonsivar: Yes sorry I clicked on the wrong, line, I meant line 142 where we have `Holder`.

EmperorOrokuSaki: That's handled in the PR 100.

subgraph/src/fleek-nfa.ts
jsonsivar: Yes I was referring to Holder here too. If that's coming in another PR that's cool then.

jsonsivar (approved):

PR #100, opened 2023-02-01
feat: replace Holder entity with Owner and handle transfer and mint

EmperorOrokuSaki:


jsonsivar:
[sc-379]

jsonsivar (approved):
LGTM. I would just keep an eye on the merge conflicts if there are once #99 is merged since there seems to be overlapping changes.

PR #101, opened 2023-02-01
chore: change top level folders

jsonsivar:
Some notes to help the review since this is a large PR but a small change (if that makes sense).

* Created top level folder for contracts and moved all contracts in there
* Split the package.json and moved contract related things into contracts/package.json but kept a root package.json for husky and other things we may need
* Updated various references to folders in husky, workflows, readmes etc.

Work being done in UI and subgraph should mostly be unaffected and merge should be smooth. For contract work we should  try to branch of this then maybe add the commits on top.


jsonsivar:
Note CI is failing, fixing that

.github/workflows/test.yml
```diff
@@ -10,6 +10,10 @@ jobs:
   test-contracts:
     runs-on: ubuntu-latest
 
+    defaults:
```
zoruka, line 13: We need to split out the ui test for other workflow, line 42 here.

.gitignore
```diff
@@ -1,20 +1,20 @@
-node_modules
+**/node_modules
 .env
 
 # hardhat
-cache
-artifacts
-deployments/hardhat
+contracts/cache
```
zoruka, line 5: Maybe we could have `.gitignore`s for each of our project folders.

README.md
```diff
@@ -21,34 +21,41 @@ You can find the wiki [here](https://github.com/fleekxyz/non-fungible-apps/wiki)
...
-- [/ui](/ui): a web application to interact with deployed contracts
 
 And after running it locally some folders may be generated:
 
-- `/artifacts`: ABIs and build info generated by Hardhat
-- `/cache`: cache info used by Hardhat
-- `/forge-cache`: cache info used by Foundry
-- `/node_modules`: all dependencies for the Node.js environment
-- `/out`: resultant ABIs for all contracts that has interactions
+- `/contracts/artifacts`: ABIs and build info generated by Hardhat
```
zoruka, line 38: Maybe we could have `README`s inside of each folder with the details of each one inside of it.

.github/workflows/test.yml
jsonsivar: The UI test is already set to working directory `ui`. Do you mean that?

zoruka: I mean this file has also the ui tests, and it is going to be triggered when has changes in the contract folder https://github.com/fleekxyz/non-fungible-apps/blob/chore/sc-281/change-top-level-folders/.github/workflows/test.yml#L46-L71

jsonsivar: I think you mean this card https://app.shortcut.com/fleekxyz/story/283/streamline-ci-to-only-do-build-tasks-associated-with-the-changes.  I was going to do it in another PR just because I wanted to research on best workflow files structure before doing that and didn't want to wait to merge this.

But basically I wanted to see what the correct structure is for monorepo. I.e., multiple files for workflows inside the top level .github folder? Or can we add .github inside each top level folder? We can filter as-is but using an external library but I think it's cleaner to do it without that and structuring it correctly. Because at the workflow level, you can filter on folders natively on GHA but not at the job level.

zoruka: I got it, we can split it after, there is no prob.

As far as I know .github needs to be in the root folder, so I think it will have multiple workflow files at the root level.

jsonsivar:
Closing this and will open a new PR from a new source branch.  Had to redo to fix the submodule error I was getting :(

PR #106, opened 2023-02-06
feat: collection and collectionOwner entities [NFA subgraph]

EmperorOrokuSaki:


EmperorOrokuSaki:
[sc-372]

jsonsivar (changes requested):
The folder structure change was merged but old things have still been committed so I would clean it up. If you're unsure we can do it together over a call but definitely can't merge this right now.

EmperorOrokuSaki:
> The folder structure change was merged but old things have still been committed so I would clean it up. If you're unsure we can do it together over a call but definitely can't merge this right now.

I can create another branch from develop for this as well.

subgraph/schema.graphql
```diff
@@ -144,6 +144,19 @@ type Owner @entity {
...
+  owners: [CollectionOwner!]
+}
+
+type CollectionOwner @entity {
+  id: Bytes! # address
+  accessGrantedBy: Bytes! #address
+  transactionHash: Bytes!
+}
+
 type Controller @entity {
```
jsonsivar, line 160: Just catching this now but we should make this either `TokenController` or `CollectionController` right?

subgraph/schema.graphql
```diff
@@ -144,6 +144,19 @@ type Owner @entity {
   tokens: [Token!] @derivedFrom(field: "owner")
 }
 
+type Collection @entity {
```
jsonsivar, line 147: Should we add collection controllers in this same PR as well? Either that are we make a new card for it. The shortrcut title had controllers accounted for but the subtasks didn't so we missed it.  But let's decide if we want it in this PR or another one. 

Above that I also created a separate card for us to handle token owner and controller changes.

subgraph/src/fleek-nfa.ts
```diff
@@ -80,6 +82,52 @@ export function handleCollectionRoleGranted(
   entity.transactionHash = event.transaction.hash;
 
   entity.save();
+
+  if (event.params.role === 0) {
```
jsonsivar, line 86: I would recommend starting to move these into separate files.  Otherwise this one file is going to get large and have merge issues etc.  Maybe we can start creating "epic" level folders here like `accessControl`, `mint`, etc and putting the handlers in the respective ones. Like this would go under access control.

subgraph/src/fleek-nfa.ts
```diff
@@ -80,6 +82,52 @@ export function handleCollectionRoleGranted(
...
+
+    let collectionOwner = CollectionOwner.load(event.params.toAddress);
+
+    if (collectionOwner) {
+      // Collection Owner already exists.
+      // Print warning log message.
+      log.warning(
+        'Although Address {} is already a collection owner, a CollectionRoleGranted event was emitted that indicated the address was granted the same role, again.',
+        [event.params.toAddress.toHexString()]
+      );
```
jsonsivar, line 99: Shouldn't this scenario still update the stored collection owner?

subgraph/src/fleek-nfa.ts
```diff
@@ -97,6 +145,13 @@ export function handleCollectionRoleRevoked(
   entity.transactionHash = event.transaction.hash;
 
   entity.save();
+
+  if (event.params.role === 0) {
```
jsonsivar, line 149: Right now we don't call this anywhere. But something to think about in the future, if we do a transfer at the collection level, which may create both the grant and revoke roles so that means both handlers will run. We should just make sure it doesn't get overwritten or deleted or anything.

subgraph/subgraph.yaml
```diff
@@ -27,12 +27,13 @@ dataSources:
...
         - TokenRoleRevoked
         - Transfer
+        - CollectionOwner
         - Token
         - Owner
         - Controller
       abis:
         - name: FleekNFA
-          file: ./abis/FleekNFA.json
+          file: ../deployments/mumbai/FleekERC721.json
```
jsonsivar, line 36: In another PR we put artifacts, and here it's deployment.  I actually think both are useful depending on what you are doing so maybe we add some docu like this:

```suggestion
          # Use this if you want to test against last deployed contracts
          #file: ../contracts/deployments/mumbai/FleekERC721.json
          # Use this if you want to test local contract changes
          file: ../contracts/artifacts/contracts/FleekERC721.sol/FleekERC721.json
```

jsonsivar (changes requested):
Main blocker to merge is merging #117 and #118 and then creating a new branch so this is clean.

Also lets make sure the ABI paths match the new folder structure.

subgraph/schema.graphql
EmperorOrokuSaki: I can change that to TokenController. Didn't see a benefit but it's a possible change.

subgraph/schema.graphql
EmperorOrokuSaki: I thought we are going to remove collection controllers per the discussion in the dev channel. Maybe @zoruka can confirm this, but since this role has no real special privileges, they are going to be removed in the ACL refactor.

subgraph/src/fleek-nfa.ts
EmperorOrokuSaki: It's the same change and the check before indicates this modification has already been applied before and the current state of the subgraph includes it.

subgraph/subgraph.yaml
EmperorOrokuSaki: I agree with adding a comment there.

subgraph/src/fleek-nfa.ts
EmperorOrokuSaki: Good point. I'll have to check if a scenario like that will result in overriding or not.

EmperorOrokuSaki:
new pr: #122

PR #109, opened 2023-02-07
feat: access point auto approval settings for tokens

EmperorOrokuSaki:
…and expose a function for changing the field later + an event.


EmperorOrokuSaki:
Tests should be updated.

contracts/test/FleekERC721.ts
```diff
@@ -1,935 +0,0 @@
-import { loadFixture } from '@nomicfoundation/hardhat-network-helpers';
```
jsonsivar, line 1: Do you guys know why this file is being shown as deleted? Look, in `develop` right now there is no file like this https://github.com/fleekxyz/non-fungible-apps/tree/develop/contracts/test but for some reason this PR is deleting it.  I just want to make sure it's not coming from some commit based off another branch than develop for example.

EmperorOrokuSaki: This branch is based on `develop`; only hardhat tests and foundry tests refactor branches were merged. It might have come from them.

jsonsivar:
Let's create a new branch from develop and cherry pick your commits for this because the commit list isn't going to be clean if we merge this. Will probably also deal with the merge conflicts.

EmperorOrokuSaki:
new pr:: #121

PR #110, opened 2023-02-07
feature: make main contract pausable

zoruka:


contracts/contracts/FleekPausable.sol
```diff
@@ -0,0 +1,134 @@
...
+error ContractIsPaused();
+error ContractIsNotPaused();
+error ContractIsNotPausable();
+error PausableIsSetTo(bool state);
+
+abstract contract FleekPausable is Initializable {
+    /**
+     * @dev Emitted when the pause is triggered by `account` and set to `isPaused`.
+     */
+    event PausedChange(bool indexed isPaused, address account);
```
EmperorOrokuSaki, line 16: Maybe we can rename this to something more descriptive like `PauseStatusChanged`

contracts/contracts/FleekPausable.sol
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * @dev Emitted when the pausable is triggered by `account` and set to `isPausable`.
+     */
+    event PausableChange(bool indexed isPausable, address account);
```
EmperorOrokuSaki, line 21: Also for this maybe we can use `PausableStatusChanged`

contracts/contracts/FleekPausable.sol
```diff
@@ -0,0 +1,134 @@
...
+
+    bool private _paused;
+    bool private _canPause; // TODO: how should we verify if the contract is pausable or not?
```
EmperorOrokuSaki, line 24: We need to think about when the contract is not pausable first. For example, when a transaction is initiated before the pause, it will revert if the contract's status changes.
At the same time, we need to consider the possibility of bad actors finding a balance of enough gas to keep the transaction pending but not enough priority gas for it to be picked up faster. They can make the contract impossible to pause externally and that's a bad position to be in.

contracts/contracts/FleekPausable.sol
```diff
@@ -0,0 +1,134 @@
...
+     */
+    modifier whenPaused() {
+        _requirePaused();
+        _;
+    }
+
+    /**
+     * @dev Returns true if the contract is paused, and false otherwise.
+     */
+    function paused() public view virtual returns (bool) {
```
EmperorOrokuSaki, line 61: maybe rename to `isPaused`

contracts/contracts/FleekPausable.sol
```diff
@@ -0,0 +1,134 @@
...
+        return _paused;
+    }
+
+    /**
+     * @dev Returns true if the contract is pausable, and false otherwise.
+     */
+    function pausable() public view virtual returns (bool) {
```
EmperorOrokuSaki, line 68: maybe rename to `isPausable`

contracts/contracts/FleekPausable.sol
```diff
@@ -0,0 +1,134 @@
...
+     */
+    function _requirePausable() internal view virtual {
+        if (!pausable()) revert ContractIsNotPausable();
+    }
+
+    /**
+     * @dev Sets the contract to be pausable or not.
+     * @param canPause true if the contract is pausable, and false otherwise.
+     */
+    function _setPausable(bool canPause) internal virtual {
```
EmperorOrokuSaki, line 97: Wondering what's the reason for the `virtual` keyword here and some of the other functions. These functions being overridable is not a good decision.

contracts/contracts/FleekPausable.sol
```diff
@@ -0,0 +1,134 @@
...
+    event PausableChange(bool indexed isPausable, address account);
+
+    bool private _paused;
+    bool private _canPause; // TODO: how should we verify if the contract is pausable or not?
+
+    /**
+     * @dev Initializes the contract in unpaused state.
+     */
+    function __FleekPausable_init() internal onlyInitializing {
+        _paused = true;
```
EmperorOrokuSaki, line 30: Doesn't this line contradict the comment on line 27?

contracts/contracts/FleekERC721.sol
```diff
@@ -527,6 +533,7 @@ contract FleekERC721 is Initializable, ERC721Upgradeable, FleekAccessControl {
      *
      * - the tokenId must be minted and valid.
      * - the sender must have the `tokenOwner` role.
+     * - the contract must be not paused.
      *
      */
     function burn(uint256 tokenId) public virtual requireTokenRole(tokenId, Roles.Owner) {
```
EmperorOrokuSaki, line 539: the `whenNotPaused` modifier is forgotten here.
Also, we need to discuss when we will use this feature and decide on what functions we want to work on when the contract is paused.

contracts/contracts/FleekERC721.sol
```diff
@@ -536,4 +543,108 @@ contract FleekERC721 is Initializable, ERC721Upgradeable, FleekAccessControl {
...
+     *
+     * - the caller should have the token role.
+     *
+     */
+    function revokeTokenRole(
+        uint256 tokenId,
+        Roles role,
+        address account
+    ) public whenNotPaused requireTokenRole(tokenId, Roles.Owner) {
+        _revokeTokenRole(tokenId, role, account);
```
EmperorOrokuSaki, line 604: Is these section going to be affected by the acl refactor?

EmperorOrokuSaki (changes requested):
I have some questions about this feature and its implementation.

I will review the tests when I understand the feature better based on the questions I have asked here in my review.

contracts/contracts/FleekERC721.sol
zoruka: Yes. I will need to fix #114 after this got merged.

contracts/contracts/FleekERC721.sol
zoruka: It is on `_beforeTokenTransfer` function that is called for mint, transfer and burn functions.

contracts/contracts/FleekPausable.sol
zoruka: Yeah, going to fix the comment

contracts/contracts/FleekPausable.sol
zoruka: That's correct, it was added by the copilot and I forgot to remove

contracts/contracts/FleekPausable.sol
zoruka: for both function names I have followed the pattern of https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/Pausable.sol, but I agree to change it

contracts/contracts/FleekPausable.sol
zoruka: Agreed, this needs to be discussed. cc @jsonsivar

jsonsivar: I'm not sure I understand the issue here yet. The _canPause is like a feature flag that only Fleek controller can call (and later can be set by a governance contract or multisig). The only time we will need to use this pause feature imo is if it's an emergency and we need to pause it to protect peoples assets. In that example like a vulnerability is discovered with our contracts or EVM e.g., I don't think it matters of any transactions initiated before the pause transaction fails right? I could be missing something here.

contracts/contracts/FleekPausable.sol
jsonsivar: Wait I think it should start in unpaused state right? Not paused but can pause would be the init state imo.

contracts/contracts/FleekPausable.sol
```diff
@@ -0,0 +1,134 @@
...
+     */
+    function _requirePausable() internal view {
+        if (!isPausable()) revert ContractIsNotPausable();
+    }
+
+    /**
+     * @dev Sets the contract to be pausable or not.
+     * @param canPause true if the contract is pausable, and false otherwise.
+     */
+    function _setPausable(bool canPause) internal {
```
jsonsivar, line 97: Should we do this like a toggle where we can switch it back too? Or create a separate function that sets canPause to false. This way its a feature flag that could be set back and forth as governance decides (and at the moment Fleek controller).

jsonsivar (changes requested):

contracts/contracts/FleekERC721.sol
```diff
@@ -527,6 +533,7 @@ contract FleekERC721 is Initializable, ERC721Upgradeable, FleekAccessControl {
      *
      * - the tokenId must be minted and valid.
      * - the sender must have the `tokenOwner` role.
+     * - the contract must be not paused.
```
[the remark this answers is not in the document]
zoruka: the modifier added to the function `_beforeTokenTransfer` that is called for mint, transfer and burn function.
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L302
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L268
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L337

contracts/contracts/FleekPausable.sol
zoruka: I think I did not get the question. The function should always invert the state? I just think this way is safer because you cannot set the flag to a state you don't want by accident.

contracts/contracts/FleekPausable.sol
jsonsivar: But in that case I think at the start we would just not deploy the contract until release moment.  I can't think of a scenario where we would want to release it but pause it at start so nobody can use it.  I noticed the Zeppelin example starts unpaused too.  Also if we leave it paused we need a second transaction just to start it.  Maybe we can make the constructor take the start pause state so we can be flexible with it?

contracts/contracts/FleekPausable.sol
jsonsivar: But we should still have a way to turn of off after. Right now once you make it pausable you can never go back?

contracts/contracts/FleekERC721.sol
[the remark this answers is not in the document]
jsonsivar: Ah gotcha. Does that make the call from here redundant then https://github.com/fleekxyz/non-fungible-apps/pull/110/files#diff-3fc5d3120a8e235b37eb66c89e4759fcead60f2b3ea2a4e5fb664892ce541b0cR183? Cuz mint also calls `_beforeTokenTransfer` right?

contracts/contracts/FleekPausable.sol
zoruka: I think we can have it unpaused by default, I don't think we need a flexible start for it tbh. I'm going to change it

contracts/contracts/FleekPausable.sol
zoruka: No, we can change it whenever we want. It just checks if you are trying to set the same value as it is, in this case it reverts.

contracts/test/foundry/FleekERC721/AccessControl.t.sol
```diff
@@ -425,4 +425,68 @@ contract Test_FleekERC721_AccessControl is Test_FleekERC721_Base {
         vm.prank(tokenOwner);
         CuT.burn(tokenId);
     }
+
```
jsonsivar, line 428: Should this go into another test file? Just wondering if access control is the best place or might be hard to tell that from the name. Maybe the pausable file or main erc721 test file.

jsonsivar (approved):
LGTM

contracts/test/foundry/FleekERC721/AccessControl.t.sol
zoruka: Here it is just testing if the address is able or not to call the function and not the function run itself, the same as other functions in this test suit.

contracts/contracts/FleekERC721.sol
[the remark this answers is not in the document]
zoruka: I think the way it is displayed by github is miss leading you to see it wrong, the modifier is not in `mint` function

jsonsivar:
Looks like merging develop into this broke CI.  I think from the error details the ABI is out of sync with the subgraph.yaml.

contracts/contracts/FleekERC721.sol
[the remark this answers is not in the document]
jsonsivar: You're right, I'll mark this as resolved.

jsonsivar:
This looks good now. Just needs approval from @EmperorOrokuSaki and it should be good to go.

PR #112, opened 2023-02-08
chore: card component

csosamorales:
- add card component
- move margin from input to label and error message


ui/src/components/card/card.styles.ts
```diff
@@ -0,0 +1,35 @@
...
+    color: '$slate12',
+    fontSize: '$xl',
+    fontWeight: '$medium',
+  });
+
+  static readonly Body = styled('div', {
+    pt: '$5',
+  });
+
+  static readonly CardText = styled('div', {
```
zoruka, line 25: Seems like we don't need `Card`here.
```suggestion
  static readonly Text = styled('div', {
```

ui/src/components/card/card.tsx
```diff
@@ -0,0 +1,69 @@
...
+        </CardStyles.CardText>
+      );
+    }
+  );
+}
+
+export namespace Card {
+  export type ContainerProps = {
+    children: React.ReactNode;
+  } & React.ComponentProps<typeof CardStyles.Container>;
```
zoruka, line 56: Seems like this is not needed, as also it change the children to have a single child.
```suggestion
  export type ContainerProps = React.ComponentProps<typeof CardStyles.Container>;

```

ui/src/components/card/card.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  export type HeadingProps = {
+    title: string;
+    leftIcon?: React.ReactNode;
+    rightIcon?: React.ReactNode;
+  } & React.ComponentProps<typeof CardStyles.Heading>;
+
+  export type BodyProps = { children: React.ReactNode } & React.ComponentProps<
+    typeof CardStyles.Body
+  >;
```
zoruka, line 66: The same from the previous comment.
```suggestion
  export type BodyProps = React.ComponentProps<
    typeof CardStyles.Body
  >;

```

ui/src/components/card/card.tsx
```diff
@@ -0,0 +1,69 @@
...
+    ({ children, ...props }, ref) => {
+      return (
+        <CardStyles.Body ref={ref} {...props}>
+          {children}
+        </CardStyles.Body>
+      );
+    }
+  );
+
+  static readonly CardText = forwardRef<HTMLDivElement, Card.CardTextProps>(
```
zoruka, line 42:
```suggestion
  static readonly Text = forwardRef<HTMLDivElement, Card.CardTextProps>(

```

ui/src/components/core/input/input.styles.ts
```diff
@@ -59,8 +59,13 @@ const styles = {
...
-export const Input = styled('input', styles);
+export const InputStyled = styled('input', styles);
 
-export const Textarea = styled('textarea', styles);
+export const InputIconStyled = styled(Icon, {
+  position: 'absolute',
+  left: '$4',
+  top: '0.9375rem',
+  color: 'slate8',
+});
```
zoruka, line 69: Should this be a variant of `Icon`?

zoruka (changes requested):
Just small change requests. Would be cool to keep page where you are adding the components to test while the project still under development, so anyone can be better aware of how you intend to use the components.

ui/src/components/core/input/input.styles.ts
csosamorales: mmm not sure, cause it's only to do this component 
image

zoruka (approved):

PR #113, opened 2023-02-08
feat: mint step 1 UI layout connect with GitHub

csosamorales:
- [x]  GH step workflow
- [x]  Repository configuration: validate not empty fields
- [x]  GH context


csosamorales:
https://user-images.githubusercontent.com/52458215/217604486-f06b6955-f43e-4800-ae40-a5dc80bee4ef.mov

ui/src/components/core/combobox/combobox.tsx
```diff
@@ -126,33 +126,36 @@ export const Combobox: React.FC<ComboboxProps> = ({
     >
       {({ open }) => (
         <>
```
zoruka, line 128: Unnecessary fragment after changes.
```suggestion

```

ui/src/theme/foundations/spacing.ts
```diff
@@ -9,8 +9,10 @@ export const spacing = {
   '3h': '0.875rem', // 14px
   4: '1rem', // 16px
   5: '1.25rem', // 20px
-  6: '1.5rem',
-  7: '1.75rem',
+  '5hh': '1.3125rem', // 21px
```
zoruka, line 12: would not be something like `5.25`?

ui/src/views/mint/github-step/github-repo-configuration.tsx
```diff
@@ -0,0 +1,130 @@
...
+    setGithubStep,
+    setRepositoryConfig,
+  } = Mint.useContext();
+  const { nextStep } = Stepper.useContext();
+  const [branchSelected, setBranchSelected] = useState(branchName);
+  const [commitHashSelected, setCommitHashSelected] = useState(commitHash);
+
+  const handlePrevStepClick = () => {
+    setGithubStep(2);
+    setRepositoryConfig('', '');
```
zoruka, line 46: Maybe clear this when the repo changes.

ui/src/views/mint/github-step/github-repo-configuration.tsx
```diff
@@ -0,0 +1,130 @@
...
+  Icon,
+  IconButton,
+  Stepper,
+} from '@/components';
+import React, { useState } from 'react';
+import { Mint } from '../mint.context';
+import { RepoRow } from './github-repository-selection';
+
+//TODO remove once it's integrated with GH login
+const branches: DropdownItem[] = [
```
zoruka, line 17: For this list we are probably going to need to fetch the branches, so we need to be aware of adding loading states for this. One scenario to verify is if github returns all branches at once or we will need to deal with some kind of pagination or search in the github api.

ui/src/views/mint/github-step/github-repo-configuration.tsx
```diff
@@ -0,0 +1,130 @@
...
+  };
+
+  const handleContinueClick = () => {
+    setRepositoryConfig(branchSelected, commitHashSelected);
+    nextStep();
+  };
+
+  return (
+    <Card.Container css={{ width: '$107h' }}>
+      <Card.Heading
```
zoruka, line 65: This is not mandatory right now, but seems like this component is going to grow when adding the remote fetching part. Probably when we got there is a good idea to split it in two components here `Header` and `Body` or something like that, so we can scope out which handlers and functions are related to each part and this would be not too much complicated since you are using context you don't need to pass props.

ui/src/views/mint/github-step/github-repository-selection.tsx
```diff
@@ -0,0 +1,156 @@
...
+  Icon,
+  IconButton,
+} from '@/components';
+import { Input } from '@/components/core/input';
+import { Separator } from '@/components/core/separator.styles';
+import React, { forwardRef, useRef, useState } from 'react';
+import { Mint } from '../mint.context';
+
+//TODO remove once it's integrated with GH login
+const repos = [
```
zoruka, line 17: Same idea from the previous component, we need to be aware with loading states and how it will be fetched from github api. Also other thing to notice is handling the error scenarios.

ui/src/views/mint/github-step/github-repository-selection.tsx
```diff
@@ -0,0 +1,156 @@
...
+    setGithubStep(1);
+  };
+
+  const handleSelectRepo = (repo: string) => {
+    setRepositoryName(repo);
+    setGithubStep(3);
+  };
+
+  return (
+    <Card.Container css={{ maxWidth: '$107h', height: '$95h', pb: '$0h' }}>
```
zoruka, line 84: Add fixed height may break the layout.

ui/src/views/mint/github-step/github-repository-selection.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  const timeOutRef = useRef<NodeJS.Timeout>();
+
+  const handleSearchChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    event.stopPropagation();
+    timeOutRef.current && clearTimeout(timeOutRef.current);
+    timeOutRef.current = setTimeout(() => {
+      setSearchValue(event.target.value);
+    }, 500);
+  };
```
zoruka, line 72: This seems to be a debounce function, right? If so, maybe it is a good idea to make from it a hook. Would be something like:
```tsx
// In the hook file
const useDebounce = <A extends any[], F extends (...args: A) => void)>(f: F, t = 500): F => {
  const timeOutRef = useRef<NodeJS.Timeout>();

  return (...args: A) => {
    timeOutRef.current && clearTimeout(timeOutRef.current);
    timeOutRef.current = setTimeout(() => {
      f(...args);
    }, t);
  };
}

// in a component
const debouncedSearch = useDebounce(
  (event: React.ChangeEvent<HTMLInputElement>) => setSearchValue(event.target.value)
);

// if you need something before line the stop propagation or you would use just `debouncedSearch`
const handleSearchChange = (event: React.ChangeEvent<HTMLInputElement>) => {
  event.stopPropagation();
  debouncedSearch(event);
}
```

zoruka: But anyway this seems like a great implementation of debounce over React, never saw it before.

ui/src/views/mint/github-step/github-repository-selection.tsx
zoruka, line 84: The same from the previous component, we may split this in different component (if the component grow up too much) to keep handlers separated and make it more readable.

ui/src/views/mint/github-step/github-repository-selection.tsx
```diff
@@ -0,0 +1,156 @@
...
+            />
+            <Input
+              leftIcon="search"
+              placeholder="Search"
+              onChange={handleSearchChange}
+            />
+          </Flex>
+          <Flex
+            css={{
+              height: '55%',
```
zoruka, line 122: Fixed height may break the layout, please double check this. Maybe `min-height` could suit as well.

ui/src/views/mint/mint-stepper.tsx
```diff
@@ -106,59 +91,62 @@ const MintStepContainer = ({ children }: MintStepContainerProps) => (
...
+            <MintStepContainer>
+              <StepperIndicatorContainer>
+                <Stepper.Indicator />
+                <Heading>Connect GitHub and select repository</Heading>
+              </StepperIndicatorContainer>
+              {/* TODO create component to handle the github connection */}
+              <GithubStep />
+            </MintStepContainer>
+          </Stepper.Step>
+          <Stepper.Step>
```
zoruka, line 109: Would be cool add a break line between steps just to make it more clear, maybe also split it in smaller components so it get like that:
```tsx
        <Stepper.Container>
          <Stepper.Step>
            <SomeStep />
          <Stepper.Step>

          <Stepper.Step>
            <SomeOtherStep />
          <Stepper.Step>

          ...

```

zoruka (approved):
Overall the code is cool, I just left some improvement comments that could be made.

The spacing file is getting a little weird. I would not add all the values needed to fit exactly the design sizes, small variations may happen. You can also edit the already existent sizes as probably they will be used really hardly ever and as much the spacing sizes are standard the UI will look better.

ui/src/views/mint/github-step/github-repo-configuration.tsx
csosamorales: Yeap, I thought about the loading state. I can add it now, or later when we add the github integration

ui/src/views/mint/github-step/github-repo-configuration.tsx
csosamorales: Ahh yeap, good catch! I was thinking to create a general card for the mint view cause I realize that we have the same style in all the card for the mint flow at least. But yeah, for now maybe I can split just this part

ui/src/theme/foundations/spacing.ts
csosamorales: removed cause I don't need it anymore

PR #114, opened 2023-02-08
refactor: clean up acl remove our redundant owner

zoruka:


contracts/contracts/FleekAccessControl.sol
```diff
@@ -5,227 +5,178 @@ pragma solidity ^0.8.7;
...
 contract FleekAccessControl is Initializable {
     using Counters for Counters.Counter;
 
-    enum Roles {
-        Owner,
-        Controller
+    /**
+     * @dev All available collection roles.
+     */
+    enum CollectionRoles {
```
EmperorOrokuSaki, line 19: Since it only has one value, maybe we can use just a string for this to make the tests more readable, too.

contracts/contracts/FleekAccessControl.sol
```diff
@@ -5,227 +5,178 @@ pragma solidity ^0.8.7;
...
-    event CollectionRoleGranted(Roles indexed role, address indexed toAddress, address byAddress);
-    event CollectionRoleRevoked(Roles indexed role, address indexed toAddress, address byAddress);
-
-    struct Role {
-        mapping(address => uint256) indexes;
-        address[] members;
+    /**
+     * @dev All available token roles.
+     */
+    enum TokenRoles {
```
EmperorOrokuSaki, line 26: Same comment as `CollectionRoles` here.

contracts/contracts/FleekAccessControl.sol
```diff
@@ -5,227 +5,178 @@ pragma solidity ^0.8.7;
...
+        Controller
     }
 
-    Counters.Counter private _collectionRolesVersion;
-    // _collectionRoles[version][role]
-    mapping(uint256 => mapping(Roles => Role)) private _collectionRoles;
+    /**
+     * @dev Emitted when a collection role is granted.
+     */
+    event TokenRoleGranted(
```
EmperorOrokuSaki, line 33: I think `TokenRoleRevoked` can be merged into this event. `TokenRoleChanged` is better probably.

contracts/contracts/FleekAccessControl.sol
```diff
@@ -5,227 +5,178 @@ pragma solidity ^0.8.7;
...
+    event TokenRoleRevoked(
+        uint256 indexed tokenId,
+        TokenRoles indexed role,
+        address indexed toAddress,
+        address byAddress
+    );
 
     /**
-     * @dev Initializes the contract by granting the `Owner` role to the deployer.
+     * @dev Emitted when token roles version is increased and all token roles are cleared and.
```
EmperorOrokuSaki, line 51: Incomplete comment.

contracts/contracts/FleekAccessControl.sol
```diff
@@ -5,227 +5,178 @@ pragma solidity ^0.8.7;
...
+     * @dev Emitted when a collection role is granted.
      */
-    modifier requireCollectionRole(Roles role) {
-        require(
-            hasCollectionRole(role, msg.sender) || hasCollectionRole(Roles.Owner, msg.sender),
-            "FleekAccessControl: must have collection role"
-        );
-        _;
-    }
+    event CollectionRoleGranted(CollectionRoles indexed role, address indexed toAddress, address byAddress);
```
EmperorOrokuSaki, line 58: Same comment as `TokenRoleChanged` here.

EmperorOrokuSaki (approved):

contracts/contracts/FleekERC721.sol
```diff
@@ -180,13 +197,13 @@ contract FleekERC721 is Initializable, ERC721Upgradeable, FleekAccessControl {
     ) internal virtual override {
         if (from != address(0) && to != address(0)) {
             // Transfer
-            _clearAllTokenRoles(tokenId, to);
+            _clearTokenRoles(tokenId);
         } else if (from == address(0)) {
             // Mint
-            _grantTokenRole(tokenId, Roles.Owner, to);
+            // TODO: set contract owner as controller
```
jsonsivar, line 203: Why is this needed?

contracts/test/hardhat/contracts/FleekERC721/token-roles.t.ts
```diff
@@ -124,86 +104,110 @@ describe('FleekERC721.TokenRoles', () => {
...
     await contract.grantTokenRole(
       tokenId,
-      Roles.Controller,
+      TokenRoles.Controller,
       otherAccount.address
     );
     await contract.transferFrom(owner.address, otherAccount.address, tokenId);
 
-    expect(await contract.getTokenRoleMembers(tokenId, 1)).to.eql([]);
+    expect(
```
jsonsivar, line 118: This comment is not needed to be done in this PR but just noting, thinking we should do a test where we add more than one controllers and then check they are all not on the list.

jsonsivar (approved):

contracts/contracts/FleekAccessControl.sol
zoruka: The thing is that it may not be a single value in the future as we add new roles. For example the verifier role we have been discussing.

contracts/contracts/FleekERC721.sol
zoruka: This needs to be set if we decided that Fleek or the contract owner should be controller for tokens, allowing to verify it. This is one thing we have not setup yet, which address should be bound as the verifier or controller that verifies each token.

contracts/contracts/FleekAccessControl.sol
EmperorOrokuSaki: that's true. for now, I think string looks better but if we are planning on adding more roles soon, then let's keep it this way.

EmperorOrokuSaki (approved):

jsonsivar (approved):
LGTM

PR #115, opened 2023-02-09
chore: mint form

csosamorales:
- [x] Layout
- [x] Context
- [x] Border radius SVG
- [x] Success status when mint
- [x] Card improvement: create component for the mint view as all the cards have the same header
- [ ] Separate body from header to have a clearer code
- [ ] debounce on inputs text

To do separate:
- [ ] Field validation: set which fields are required
- [ ] Color picker component
- [ ] Switch component

Nice to have
- [ ] wallet connection, just the button
- [ ] real mint function

https://user-images.githubusercontent.com/52458215/218130313-f07a718e-a999-47ff-a054-0dbbff4bac00.mov


csosamorales:
[sc-331]
[sc-330]

ui/src/views/mint/form-step/fields/logo/index.ts
```diff
@@ -0,0 +1 @@
+export * from './logo-field';
```
zoruka, line 1: The folder `logo` should not be `logo-field` also to keep the pattern?

ui/src/views/mint/form-step/fields/verify-nfa-field.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  const handleVerifyNFAChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setVerifyNFA(e.target.checked);
+  };
+
+  return (
+    <Card.Text css={{ p: '$4', textAlign: 'left' }}>
+      {/* TODO replace for grid */}
+      <Flex css={{ gap: '$10' }}>
+        <Grid css={{ rowGap: '$1h' }}>
```
zoruka, line 15: It seems like here could be a single grid, maybe the `Card.Text` could be used as a grid in this case.

ui/src/views/mint/form-step/form.utils.ts
```diff
@@ -0,0 +1,29 @@
+//TODO create env variable
+const DEFAULT_MAX_FILE_SIZE = 10; // in KB
+
+// The file size must be capped to a size that the contract can handle
```
zoruka, line 4: Could keep the function comment in the same pattern of the others.

ui/src/views/mint/form-step/mint-form.tsx
```diff
@@ -0,0 +1,36 @@
...
+export const FormStep = () => {
+  const { prevStep, nextStep } = Stepper.useContext();
+  const { appName, appDescription, domain } = Mint.useContext();
+
+  return (
+    <Card.Container css={{ width: '$107h' }}>
+      <MintCardHeader title="NFA Details" onClickBack={prevStep} />
+      <Card.Body>
+        <Grid css={{ rowGap: '$6' }}>
+          <Grid css={{ rowGap: '$4' }}>
```
zoruka, line 17: These elements could be a single grid using grid template and areas.

zoruka (approved):
Great work!

ui/src/views/mint/form-step/mint-form.tsx
csosamorales: will do!

PR #117, opened 2023-02-10
test: init subgraph unit tests

EmperorOrokuSaki:


subgraph/yarn.lock
```diff
@@ -1445,9 +1445,9 @@ glob@^7.1.3:
     once "^1.3.0"
     path-is-absolute "^1.0.0"
 
-"gluegun@git+https://github.com/edgeandnode/gluegun.git#v4.3.1-pin-colors-dep":
+"gluegun@https://github.com/edgeandnode/gluegun#v4.3.1-pin-colors-dep":
```
jsonsivar, line 1448: Is this an upgrade we did manually?

jsonsivar (approved):
LGTM. But let's make sure we run the codegen, test and build commands before  we push sinc we had the build command fail. Maybe we can add those to husky.

jsonsivar:
Also, added @zoruka as a reviewer here to at least review the contract changes b4 merge.

zoruka (approved):

subgraph/yarn.lock
EmperorOrokuSaki: It happened automatically.

EmperorOrokuSaki:
> LGTM. But let's make sure we run the codegen, test and build commands before we push sinc we had the build command fail. Maybe we can add those to husky.

commit [8a35751](https://github.com/fleekxyz/non-fungible-apps/pull/117/commits/8a357512e24250998ed44e01e31bd5495fc7733a) already includes that.

jsonsivar:
> > LGTM. But let's make sure we run the codegen, test and build commands before we push sinc we had the build command fail. Maybe we can add those to husky.
> 
> commit [8a35751](https://github.com/fleekxyz/non-fungible-apps/pull/117/commits/8a357512e24250998ed44e01e31bd5495fc7733a) already includes that.

Yes because we chatted on it so I meant for future PRs.  Also the comment on adding to husky still applies right? I mean we don't have to add it to this PR but at some point we should?

EmperorOrokuSaki:
> > > LGTM. But let's make sure we run the codegen, test and build commands before we push sinc we had the build command fail. Maybe we can add those to husky.
> > 
> > 
> > commit [8a35751](https://github.com/fleekxyz/non-fungible-apps/pull/117/commits/8a357512e24250998ed44e01e31bd5495fc7733a) already includes that.
> 
> Yes because we chatted on it so I meant for future PRs. Also the comment on adding to husky still applies right? I mean we don't have to add it to this PR but at some point we should?

Yes, we should. I think this can be a part of the deployment task for prs that target `main`. We can continue to discuss this in the dev channel since the PR is closed

PR #118, opened 2023-02-10
feat: add subgraph unit tests to github action

EmperorOrokuSaki:
Runs subgraph unit tests on PRs to main


.github/workflows/subgraph.yml
```diff
@@ -0,0 +1,35 @@
+name: Subgraph
```
jsonsivar, line 1: I changed the base branch to the branch from #117 but I'm still seeing the diff from that PR.  Technically this file and the next file should be the only changes in this PR right?

.github/workflows/subgraph.yml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  pull_request:
+    branches:    
+      - main
```
jsonsivar, line 6: Should we start with `develop`?

.github/workflows/subgraph.yml
```diff
@@ -0,0 +1,35 @@
...
+
```
jsonsivar, line 7: We should also filter on changes to the subgraph directory.
```suggestion

  push:
    paths:
      - 'subgraph/**'
```

.github/workflows/subgraph.yml
```diff
@@ -0,0 +1,35 @@
...
+          node-version: ${{ matrix.node-version }}
+      
+      - name: Install dependencies
+        run: yarn --ignore-scripts
+
+      - name: Run codegen
+        run: yarn codegen
+          
+      - name: Test Subgraph
+        run: yarn test
```
jsonsivar, line 35: What about the `build` command?

jsonsivar (changes requested):

.github/workflows/subgraph.yml
jsonsivar: We can change it back to `develop` once we merge the other PR I guess. But should make sure the diff is right because I can't think of why it would show these commits/diff here.

.github/workflows/test.yml
```diff
@@ -9,11 +9,11 @@ on:
 jobs:
   test-contracts:
     runs-on: ubuntu-latest
-
+
```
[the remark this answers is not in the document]
EmperorOrokuSaki: yes. VS Code adds tabs for indentation when you add a new line.

.github/workflows/subgraph.yml
EmperorOrokuSaki: Changed the base back to `develop` but still see the diffs. Since 117 was merged into this, this branch carries those changes, too.
I think we can go ahead and merge since the only other changes except for the ones that were from 117 are the GH action file changes.
But I can cherry-pick if that's what you prefer.

.github/workflows/subgraph.yml
EmperorOrokuSaki: We discussed it in the call and agreed with `main` if I remember correctly. The ABI path is going to point to the Mumbai deployments, and not deploying a new version for every subgraph change that are in PRs made to `develop` will cause problems.
I don't see an advantage to adding `develop` here; Especially since Subgraph changes should be one of the last things to be done before a new release.

.github/workflows/subgraph.yml
EmperorOrokuSaki: I think my other command is relevant to this as well: https://github.com/fleekxyz/non-fungible-apps/pull/118#discussion_r1103052509

.github/workflows/subgraph.yml
EmperorOrokuSaki: Only needed for deployments to Hosted Service.
This PR contains commands to run Subgraph unit tests.

.github/workflows/test.yml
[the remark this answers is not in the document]
jsonsivar: I meant shoudn't.  Typically we don't have empty tabs like this no? I don't see this empty tabs anywhere else.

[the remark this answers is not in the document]
EmperorOrokuSaki: It adds them and I manually delete them lol. I'll remove this one in the new branch as well.

.github/workflows/subgraph.yml
jsonsivar: Yeah let's cherry pick into new branch based off develop once 117 is merged so that it's clean.  I know it's an extra step but we know if we do it clean the diff shouldn't show and this will help keep things very clean :pray:

.github/workflows/subgraph.yml
jsonsivar: How so? Even if we do `main` this applies.  This is about filtering on only changes to this folder. So this wont run on changes to other folders. But perhaps to improve we can add contracts here too so that if there is a contract change that breaks subgraph we catch it. But regardless I think this is not affected by what we decide for the other comment.

.github/workflows/subgraph.yml
jsonsivar: But we just saw in the other PR that a change broke the build command.  Why would we want to submit changes we know break the build command?

.github/workflows/subgraph.yml
EmperorOrokuSaki: Every release to `main` will most likely have a code change and that code change is going to result in a new ABI, a new deployed instance of the contract and it makes sense for subgraph unit tests to require changes as well. But if we want to add `develop`, then this comment to add the path makes a difference so that this github action does not run when there's no change.

.github/workflows/subgraph.yml
EmperorOrokuSaki: I think it's just that especially now that we are going to run the gh action for develop prs too, some contract change that changes the event ABIs can break the subgraph build command.

But as long as the default value for the path to the ABI is deployments, that shouldn't make a difference so I'll add the build command here too. (It will fail for the next PR because the ABI of the deployed contract will be different from the changes in the subgraph config files since we have not yet deployed the new version.)

EmperorOrokuSaki:
Will close this PR.
New PR: #120

PR #119, opened 2023-02-10
feat: expose a function that returns all non-nested values of an App

EmperorOrokuSaki:


contracts/test/hardhat/contracts/FleekERC721/get-token.t.ts
```diff
@@ -0,0 +1,28 @@
...
+import { expect } from 'chai';
+import { TestConstants, Fixtures, parseTokenURI } from './helpers';
+import { ethers } from 'hardhat';
+const { MintParams } = TestConstants;
+
+describe('FleekERC721.GetToken', () => {
+  let fixture: Awaited<ReturnType<typeof Fixtures.withMint>>;
+
+  before(async () => {
+    fixture = await loadFixture(Fixtures.withMint);
```
zoruka, line 11: maybe here would be better be `beforeEach` since if we add more tests, if one of them change the fixture the change will be kept to other tests.

contracts/test/foundry/FleekERC721/GetToken.t.sol
```diff
@@ -0,0 +1,73 @@
...
+        vm.prank(address(1));
+        CuT.getToken(tokenId);
+        vm.prank(address(2));
+        CuT.getToken(tokenId);
+        vm.prank(address(3));
+        CuT.getToken(tokenId);
+    }
+
+    function testFail_tokenURIForNonExistentId() public view {
+        CuT.getToken(1);
```
zoruka, line 71: maybe here could use `expectRevertWithInvalidTokenId()` instead of `testFail`.

zoruka (approved):
LGTM. Just left comments about small improvements.

contracts/contracts/FleekERC721.sol
```diff
@@ -57,7 +57,6 @@ contract FleekERC721 is Initializable, ERC721Upgradeable, FleekAccessControl {
         string ENS; // ENS ID
         uint256 currentBuild; // The current build number (Increments by one with each change, starts at zero)
         mapping(uint256 => Build) builds; // Mapping to build details for each build number
-        string[] accessPoints; // List of app AccessPoint
```
jsonsivar, line 60: Created this card so we make sure we create the subgraph updates for it: https://app.shortcut.com/fleekxyz/story/458/create-events-and-handlers-for-ap.

jsonsivar (approved):
LGTM

PR #126, opened 2023-02-13
Fix tests

schmidsi:


subgraph/src/fleek-nfa.ts
```diff
@@ -101,7 +101,7 @@ export function handleNewBuild(event: NewBuildEvent): void {
   let entity = new NewBuild(
     event.transaction.hash.concatI32(event.logIndex.toI32())
   );
-  entity.token = event.params.tokenId;
+  entity.token = event.params.token;
```
schmidsi, line 104: Seems like the ABI changed, so I had to fix this

subgraph/subgraph.yaml
```diff
@@ -39,17 +39,17 @@ dataSources:
...
-        - event: NewTokenDescription(indexed uint256,string,indexed address)
+        - event: NewTokenDescription(indexed uint256,indexed string,indexed address)
           handler: handleNewTokenDescription
-        - event: NewTokenENS(indexed uint256,string,indexed address)
+        - event: NewTokenENS(indexed uint256,indexed string,indexed address)
           handler: handleNewTokenENS
-        - event: NewTokenExternalURL(indexed uint256,string,indexed address)
+        - event: NewTokenExternalURL(indexed uint256,indexed string,indexed address)
           handler: handleNewTokenExternalURL
-        - event: NewTokenLogo(indexed uint256,string,indexed address)
```
schmidsi, line 50: Also new (old?) event names

subgraph/tests/utils.ts
```diff
@@ -162,7 +168,12 @@ export function createNewBuildEvent(
     new ethereum.EventParam('token', ethereum.Value.fromUnsignedBigInt(token))
   );
   newBuildEvent.parameters.push(
-    new ethereum.EventParam('commitHash', ethereum.Value.fromString(commitHash))
+    new ethereum.EventParam(
+      'commitHash',
+      ethereum.Value.fromBytes(
```
schmidsi, line 173: Here is where the magic happens: Since the event parameter is seen as "bytes", the trick is to actually create a Bytes value. The naming of the function "fromBytes" as a bit confusing but it also sets the type of the value to "BYTES":

```
    static fromBytes(bytes: Bytes): Value {
      return new Value(ValueKind.BYTES, changetype<u32>(bytes))
    }
```

As seen here: https://github.com/graphprotocol/graph-ts/blob/main/chain/ethereum.ts#L319

schmidsi:
The question remains if it makes sense to store `indexed string` as Bytes. I think there are some considerations.

schmidsi:
Here is some context: https://docs.soliditylang.org/en/v0.8.6/abi-spec.html?highlight=indexed#encoding-of-indexed-event-parameters

jsonsivar:
We turned off indexing for the string parameters anyways so this became a non-issue. We will close this for now and reopen if we need this technique in the future. Also added it to our discussion around matchstick testing so we can refer to it if needed https://github.com/fleekxyz/non-fungible-apps/discussions/168.

PR #131, opened 2023-02-16
feat: subgraph reverse lookup (gitRepo -> NFA)

EmperorOrokuSaki:


subgraph/schema.graphql
```diff
@@ -38,6 +38,25 @@ type CollectionRoleRevoked @entity(immutable: true) {
   transactionHash: Bytes!
 }
 
+type NewMint @entity(immutable: true) {
```
jsonsivar, line 41: Why do we need a new entitty? Can't we add these properties to the existing token entitty? I think we are still using entities the wrong way?

subgraph/src/fleek-nfa.ts
```diff
@@ -153,6 +156,89 @@ export function handleCollectionRoleRevoked(
...
+  let gitRepository = event.params.gitRepository;
+  let commitHash = event.params.commitHash;
+  let logo = event.params.logo;
+  let color = event.params.color;
+  let accessPointAutoApprovalSettings =
+    event.params.accessPointAutoApprovalSettings;
+  let tokenId = event.params.tokenId;
+  let ownerAddress = event.params.owner;
+
+  newMintEntity.tokenId = tokenId;
```
jsonsivar, line 177: As per the comment above, we can remove the mint entity and just go directly from mint event to updating the token entity right?

subgraph/subgraph.yaml
```diff
@@ -6,9 +6,9 @@ dataSources:
     name: FleekNFA
     network: mumbai
     source:
-      address: "0x34F21E970A7cd383eE429aDB5ed57bbc40ea2B57"
+      address: "0x113169680A417e4dB4AD11832486FF317fB4CB36"
```
jsonsivar, line 9: Can we confirm this is the proxy address? cc @zoruka

subgraph/schema.graphql
```diff
@@ -160,3 +188,8 @@ type CollectionOwner @entity {
   accessGrantedBy: Bytes! #address
   transactionHash: Bytes!
 }
+
+type GitRepository @entity {
```
jsonsivar, line 192: Should we add a test or example query to query this via the reverse lookup?

jsonsivar (changes requested):
This doesn't appear to be branched from develop. Also I don't think we need the intermediate mint entity.

subgraph/schema.graphql
EmperorOrokuSaki: No, this is the `NewMint` event entity and essential to handling the event and querying it in the future like other events in the schema file. `NewTokenName`, `Approval`, `GrantCollectionRole` all of these are also event entities and they are all immutable.

subgraph/src/fleek-nfa.ts
EmperorOrokuSaki: Yes, but why wouldn't we want to keep the record of the first mint? The best practice is to keep records of all events from the contract in subgraph.

subgraph/subgraph.yaml
EmperorOrokuSaki: It's not. should change.

EmperorOrokuSaki:
> This doesn't appear to be branched from develop. Also I don't think we need the intermediate mint entity.

It is branched from `develop` (GitHub insights also shows that). It contains code from auto approval because initially, it was a different branch that was accidentally branched from that but I merged auto approval locally to develop on my machine and created a new branch based on that, cherry-picked, and pushed (I think this was also briefly discussed on Discord dev channel before)

jsonsivar (approved):
LGTM

PR #132, opened 2023-02-17
feat: integrate firebase for github login

csosamorales:
[sc-409]


ui/src/views/mint/github-step/steps/github-repository-selection/repositories-list.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const handleSelectRepo = (repo: Repo) => {
+    setRepositoryName(repo);
+    setGithubStep(3);
+    setRepositoryConfig({} as DropdownItem, '');
+  };
+
+  const { data: dataRepositories, status } = useQuery(
+    `fetchRepos${selectedUserOrg.value}`,
+    async () => fetchRepos(selectedUserOrg.value)
+  );
```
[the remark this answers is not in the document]
zoruka: Also would be cool for this if we can add a type for `data` in these calls, so it keeps easier to use in the app instead of dealing with `any`s.

ui/src/views/mint/github-step/steps/github-repository-selection/repositories-list.tsx
```diff
@@ -0,0 +1,110 @@
...
+                  variant="outline"
+                  css={{ py: '$1', height: '$5', borderRadius: '$md' }}
+                  onClick={() => handleSelectRepo(repo)}
+                >
+                  Use for NFA
+                </Button>
+              }
+            />
+            {index < length - 1 && <Separator />}
+          </React.Fragment>
```
zoruka, line 103: Maybe this could be moved out for a smaller component with the handlers out of the render.

zoruka (changes requested):

ui/src/components/core/index.ts
```diff
@@ -2,3 +2,5 @@ export * from './button';
 export * from './combobox';
 export * from './icon';
 export * from './input';
+export * from './avatar';
+export * from './separator.styles';
```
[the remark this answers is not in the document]
csosamorales: 🤦🏼‍♀️ yes

ui/src/store/features/github/async-thunk/github-api.tsx
```diff
@@ -0,0 +1,76 @@
+import { ComboboxItem, DropdownItem } from '@/components';
+import { githubActions, RootState } from '@/store';
+import { useGithub } from '@/views/mint/github-step/steps/use-github';
+import { Repo } from '@/views/mint/mint.context';
```
zoruka, line 4: It seems not correct import things from the view in redux files the direction should be the opposite.

ui/src/store/features/github/async-thunk/github-api.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useGithub } from '@/views/mint/github-step/steps/use-github';
+import { Repo } from '@/views/mint/mint.context';
+import { createAsyncThunk } from '@reduxjs/toolkit';
+
+export const fetchRepositoriesThunk = createAsyncThunk(
+  'github/fetchRepositories',
+  async (url: string, { dispatch, getState }) => {
+    debugger;
+    if ((getState() as RootState).github.queryLoading === 'loading') return;
+    const { fetchRepos } = useGithub();
```
zoruka, line 12: This seems circular to be a circular dependency. We should not have hooks inside an async thunk. Hooks should only be used for React components.

ui/src/store/features/github/github-slice.ts
```diff
@@ -0,0 +1,96 @@
...
+  export type State = 'disconnected' | 'loading' | 'connected';
+
+  export type QueryLoading = 'idle' | 'loading' | 'failed' | 'success';
+
+  export type UserAndOrganizations = Array<ComboboxItem>;
+
+  export type Repositoires = Array<Repo>;
+
+  export type Branches = Array<DropdownItem>;
+}
```
zoruka, line 20: `Repo` type from view should be moved to here.

ui/src/views/mint/github-step/steps/github-connect/use-firebase.ts
```diff
@@ -0,0 +1,63 @@
+import { useCallback } from 'react';
```
zoruka, line 1: I think this is going to be removed, right?

ui/src/views/mint/github-step/steps/use-github.ts
```diff
@@ -0,0 +1,69 @@
+import { DropdownItem } from '@/components';
+import { useGithubStore } from '@/store';
+import { Octokit } from 'octokit';
+
+export const useGithub = () => {
```
zoruka, line 5: I recommend move this hook to the async thunk file and change it from hook to basic functions, you are able to retrieve the token directly in redux state and this would remove the circular dependency.

zoruka (commented):
Some improvement comments

ui/src/store/features/github/async-thunk/fetch-user-organizations.ts
```diff
@@ -0,0 +1,38 @@
...
+      .github;
+
+    if (queryUserAndOrganizations === 'loading') return;
+
+    try {
+      dispatch(githubActions.setQueryUserState('loading'));
+
+      const githubClient = new GithubClient(token);
+      const user = await githubClient.fetchUser();
+      const orgs = await githubClient.fetchOrgs();
```
zoruka, line 20: For situations where the requests are complementary and not dependent on each other we can use `Promise.all` to trigger them concurrently and improve the response time.
```suggestion
const [user, orgs] = await Promise.all([githubClient.fetchUser(), githubClient.fetchOrgs()]);
```

zoruka (approved):
Good work! Just an improvement comment.

PR #135, opened 2023-02-20
feat: metadata update and token changes on subgraph

EmperorOrokuSaki:


subgraph/subgraph.yaml
```diff
@@ -6,9 +6,9 @@ dataSources:
     name: FleekNFA
     network: mumbai
     source:
-      address: "0x34F21E970A7cd383eE429aDB5ed57bbc40ea2B57"
+      address: "0x113169680A417e4dB4AD11832486FF317fB4CB36"
```
jsonsivar, line 9: Should this be the proxy address instead? cc @zoruka

jsonsivar (approved):
LGTM.  Made a card for improvements later on we can discuss that on the card or scrums :+1: https://app.shortcut.com/fleekxyz/story/578/subgraph-refactor-metadata-update-handlers

subgraph/subgraph.yaml
EmperorOrokuSaki: changed to proxy.

PR #137, opened 2023-02-21
chore: remove counters library from the contract code.

EmperorOrokuSaki:
Discussion: #129


contracts/contracts/FleekERC721.sol
```diff
@@ -133,8 +133,8 @@ contract FleekERC721 is Initializable, ERC721Upgradeable, FleekAccessControl, Fl
 
         // The mint interaction is considered to be the first build of the site. Updates from now on all increment the currentBuild by one and update the mapping.
         app.currentBuild = 0;
-        app.builds[0] = Build(commitHash, gitRepository);
-
+        app.builds[0] = Build(commitHash, gitRepository);    
+
```
jsonsivar, line 137: Just some whitespace to clear

jsonsivar (approved):
LGTM

contracts/contracts/FleekERC721.sol
```diff
@@ -86,6 +84,7 @@ contract FleekERC721 is Initializable, ERC721Upgradeable, FleekAccessControl, Fl
     function initialize(string memory _name, string memory _symbol) public initializer {
         __ERC721_init(_name, _symbol);
         __FleekAccessControl_init();
+        _appIds = 0;
```
zoruka, line 87: Do we need to initialize also in the AccessControl side?

zoruka (approved):

contracts/contracts/FleekERC721.sol
EmperorOrokuSaki: In mappings, the default value is always zero.

PR #140, opened 2023-02-24
feat: sc 121 - serverless boilerplate

jsonsivar:


serverless/.env.development
```diff
@@ -0,0 +1 @@
+# Add your environment variables here
```
zoruka, line 1: Should this file be git ignored?

serverless/.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
zoruka, line 4: `json` files usually should not have comments 🤔

serverless/lint-staged.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
zoruka, line 1: Seems like this config is overlapping the config on `.husky/lint-staged.config.json` maybe we need to setup single rules for the entire project although with the root `.eslintrc` file. The same applies to some packages that are installed for `serverless` project that seems like it should be over the root project like the packages of `@commitlint`.

zoruka (commented):
Left some comments about the package structure and I will be trying to run it locally.

serverless/.vscode/settings.json
```diff
@@ -0,0 +1,29 @@
...
+    "package-lock.json": true
+  },
+  "editor.defaultFormatter": "dbaeumer.vscode-eslint",
+  "editor.formatOnSave": false,
+  "editor.codeActionsOnSave": [
+    "source.addMissingImports",
+    "source.fixAll.eslint"
+  ],
+  "jest.autoRun": {
+    "watch": false // Start the jest with the watch flag
```
EmperorOrokuSaki, line 14: json comments also here

serverless/README.md
```diff
@@ -0,0 +1,87 @@
...
+### Requirements
+
+- Node.js 14+ and npm
+
+### Getting started
+
+Run the following command on your local environment:
+
+```
+git clone --depth=1 https://github.com/ixartz/Serverless-Boilerplate-Express-TypeScript.git my-project-name
```
EmperorOrokuSaki, line 48: do we need to include this step?

EmperorOrokuSaki (changes requested):
I think there's an issue with the dependencies. When I tried to install the dependencies, yarn gave me this error:

```

error concat-text-webpack-plugin@0.2.1: The engine "node" is incompatible with this module. Expected version ">=8 <=16". Got "18.13.0"
error Found incompatible module.
```

I think we should change the dependencies so that newer versions of node can be used, too.

serverless/.env.development
jsonsivar: It was meant to be a template for .env with needed values so I fixed this by giving default values.

serverless/.vscode/settings.json
jsonsivar: Fixed in [62680f9](https://github.com/fleekxyz/non-fungible-apps/pull/140/commits/62680f92bc6a10e769ee4314edfaafa191ee4194).

serverless/README.md
jsonsivar: Fixed in [62680f9](https://github.com/fleekxyz/non-fungible-apps/pull/140/commits/62680f92bc6a10e769ee4314edfaafa191ee4194).

serverless/.vscode/tasks.json
```diff
@@ -0,0 +1,21 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558
+  // for the documentation about the tasks.json format
```
zoruka, line 3: Comments here too

serverless/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+After cloning the repo, ensure you run `yarn` in the root directory. After that, `cd` into the `serverless` directory and alsy run `yarn`.
+
+If you are deploying, make sure you have your AWS credentials set to environment variables or have setup AWS credentials using the AWS CLI.
+
+Basically, these values need to be set:
+```
+export AWS_ACCESS_KEY_ID=value
+export AWS_SECRET_ACCESS_KEY=value
+export AWS_SESSION_TOKEN=value
```
zoruka, line 17: are we able to use a `.env` file?

zoruka (commented):
I would recommend using `this-pattern-of-name.ts` instead using camel case to keep consistency between other projects in the repo.

serverless/README.md
jsonsivar: I don't think we can use `.env` for that.  Since it is `serverless` and not a `node` environment, it doesn't pick it up.  But I think the proper way is to use the AWS CLI as per these instructions: https://www.serverless.com/framework/docs/providers/aws/guide/credentials/.

jsonsivar:
> I would recommend using `this-pattern-of-name.ts` instead using camel case to keep consistency between other projects in the repo.

Agreed and will update.  Is there a rule we can add in the eslint for this?

jsonsivar:
> I would recommend using `this-pattern-of-name.ts` instead using camel case to keep consistency between other projects in the repo.

Updated to be "kebab-case" (TIL). Also added `eslint` rule for it.

zoruka (approved):

EmperorOrokuSaki (approved):
looks good to me

csosamorales (approved):

PR #147, opened 2023-02-28
chore: UI verification step separate

csosamorales:
- Verify NFA on a separate step
- Move form step inside NFA Step folder
- New Switch component [sc-446]


zoruka (approved):
LGTM. It is a good idea double check the fixed `width` and `height` used overall, it may break the layout as the page size changes.

PR #148, opened 2023-03-01
feat: add verifier role

zoruka:


contracts/contracts/FleekERC721.sol
```diff
@@ -305,7 +304,7 @@ contract FleekERC721 is Initializable, ERC721Upgradeable, FleekAccessControl, Fl
     ) public virtual requireTokenOwner(tokenId) {
         _requireMinted(tokenId);
         _apps[tokenId].accessPointAutoApproval = _apAutoApproval;
-        emit MetadataUpdate(tokenId, 'accessPointAutoApproval', _apAutoApproval, msg.sender);
+        emit MetadataUpdate(tokenId, "accessPointAutoApproval", _apAutoApproval, msg.sender);
```
EmperorOrokuSaki, line 307: is there a reason to change quotations here?

EmperorOrokuSaki (approved):

contracts/contracts/FleekERC721.sol
zoruka: No, is just the auto formatting 😆

contracts/contracts/FleekERC721.sol
```diff
@@ -624,13 +623,13 @@ contract FleekERC721 is Initializable, ERC721Upgradeable, FleekAccessControl, Fl
...
      * - the AP must exist.
-     * - the sender must have the token controller role.
+     * - the sender must have collection verifier role.
      *
      */
     function setAccessPointNameVerify(
         string memory apName,
         bool verified
-    ) public requireAP(apName) requireTokenRole(_accessPoints[apName].tokenId, TokenRoles.Controller) {
+    ) public requireAP(apName) requireCollectionRole(CollectionRoles.Verifier) {
```
jsonsivar, line 632: I created this card so we can add the verified field to the main NFA and it's setter would also need a require like this.https://app.shortcut.com/fleekxyz/story/657/contracts-add-verified-field-and-allow-verifier-role-to-edit-it

contracts/test/hardhat/contracts/FleekERC721/collection-roles.t.ts
```diff
@@ -175,4 +175,39 @@ describe('FleekERC721.CollectionRoles', () => {
...
+    );
+
+    await contract.addAccessPoint(0, 'random.com');
+
+    await expect(
+      contract
+        .connect(otherAccount)
+        .setAccessPointContentVerify('random.com', true)
+    )
+      .to.be.revertedWithCustomError(contract, Errors.MustHaveCollectionRole)
```
jsonsivar, line 202: Maybe we can expect rever with calling it by token owner too to make sure they should never be able to set it.

jsonsivar (approved):

contracts/test/hardhat/contracts/FleekERC721/collection-roles.t.ts
zoruka: Changed the token owner to be `otherAccount` since `owner` also has the verifier role

PR #151, opened 2023-03-02
feat: erc interface and split out access points to single module

zoruka:


jsonsivar (approved):
LGTM. There is one merge conflict in the AP verify function. We should require that both they are valid verifiers and also that they are a token controller right?

zoruka:
> LGTM. There is one merge conflict in the AP verify function. We should require that both they are valid verifiers and also that they are a token controller right?

The verifier don't need to be controller anymore. We are also adding in further PR #164 about setting the verifier that will be able to verify, so it doesn't require to be also controller.

jsonsivar (approved):

PR #152, opened 2023-03-02
chore: UI setup subgraph querying

csosamorales:
- Update ABI conctract to match with subgraph
- Added Apollo and Graph Client
- Setup subgraph
- Basic pagination [sc-611]
- Add 'create-ap' route and button on NFA card [sc-644]


zoruka (changes requested):
Needs to ignore generated files and add how to generate them in the README.

ui/src/views/home/nfa-list/nfa-list.tsx
```diff
@@ -0,0 +1,91 @@
...
+              <Card.Body css={{ display: 'flex', flexDirection: 'column' }}>
+                <a
+                  target="_blank"
+                  href={`https://testnets.opensea.io/assets/mumbai/${FleekERC721.address}/${mint.tokenId}`}
+                >
+                  <u>Open NFA on Opensea</u>
+                </a>
+                <Link to={`/create-ap/${mint.tokenId}`}>Create AP</Link>
+              </Card.Body>
+            </Card.Container>
```
zoruka, line 83: Would be cool split this in a smaller component in future.

csosamorales: Will do, this is just as an example since we don't have the designs for this

.github/workflows/test.yml
```diff
@@ -64,6 +64,9 @@ jobs:
       - name: Install Dependencies
         run: yarn --ignore-scripts
 
+      - name: Create .graphclient folder
```
EmperorOrokuSaki, line 67: we should make sure this PR pulls from develop before merging because this might override the new workflow files.

ui/queries.graphql
```diff
@@ -0,0 +1,15 @@
+query lastTenMints($skip: Int) {
+  newMints(first: 2, skip: $skip, orderDirection: desc, orderBy: tokenId) {
```
EmperorOrokuSaki, line 2: This only returns two NewMint values. The name is misleading.

ui/src/integrations/ethereum/contracts/FleekERC721.json
```diff
@@ -1,9 +1,88 @@
 {
-  "timestamp": "2/15/2023, 12:43:39 PM",
```
EmperorOrokuSaki, line 2: I have a question about this file. I think we have an identical version of this in `contracts` so is it necessary to have it here as well?

EmperorOrokuSaki (approved):

ui/src/integrations/ethereum/contracts/FleekERC721.json
csosamorales: Yeah, we need it for the integration

ui/.graphclientrc.yml
```diff
@@ -0,0 +1,9 @@
+# .graphclientrc.yml
+sources:
+  - name: FleekNFA
+    handler:
+      graphql:
+        endpoint: https://api.thegraph.com/subgraphs/name/emperororokusaki/flk-test-subgraph #replace for nfa subgraph
+
+documents:
+  - ./queries.graphql
```
zoruka, line 9: Maybe is better have a folder for graphql files and get all files
```suggestion
  - ./graphql/*/*.graphql
```

zoruka (approved):

PR #153, opened 2023-03-03
chore: add / update workflows to do folder filtering

EmperorOrokuSaki:


.github/workflows/subgraph.yml
```diff
@@ -6,7 +6,6 @@ on:
       - main
 
 jobs:
-
```
csosamorales, line 9: Doesn't you need to install the dependencies to run the tests?

EmperorOrokuSaki: It does install the dependencies at line 30

csosamorales (approved):

.github/workflows/ui.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install Dependencies
+        run: yarn --ignore-scripts
+- name: Create .graphclient folder
```
zoruka, line 32: This line is not off?

EmperorOrokuSaki: it is

zoruka (approved):

.github/workflows/subgraph.yml
```diff
@@ -6,7 +6,6 @@ on:
       - main
```
jsonsivar, line 7: Do we need the folder filter here?

EmperorOrokuSaki: Since we are updating subgraph tests before every release, doing folder filtering raises incompatibility issues between the outdated tests and the new mapping Assembly Script file.
Maybe having folder filtering and the current condition together can work, but still I think for all merges to main, the workflow should run.

jsonsivar (approved):
LGTM :pray:

PR #156, opened 2023-03-03
feature: UI fetch mint price on contract and send correct value

zoruka:


zoruka:
Creating a new branch to cherry pick and remove old commit

PR #161, opened 2023-03-07
chore: separate handlers by event type [subgraph]

EmperorOrokuSaki:


subgraph/src/fleek-nfa.ts
```diff
@@ -1,602 +1,7 @@
...
-    accessPointEntity.save();
-  } else {
-    // Unknown access point
-    log.error(
-      'Unable to handle ChangeAccessPointContentVerify. Unknown access point. Verified: {}, AccessPoint: {}',
-      [event.params.verified.toString(), event.params.apName]
-    );
-  }
-}
+import './access-control';
```
jsonsivar, line 1: I think we need to export this no?

jsonsivar (changes requested):

jsonsivar (approved):

PR #162, opened 2023-03-07
feat: add handler for burns to the transfer handler [subgraph]

EmperorOrokuSaki:


subgraph/src/fleek-nfa.ts
```diff
@@ -423,23 +426,29 @@ export function handleTransfer(event: TransferEvent): void {
   }
```
jsonsivar, line 426: Do we need to save the owner entity above here?

subgraph/src/fleek-nfa.ts
```diff
@@ -423,23 +426,29 @@ export function handleTransfer(event: TransferEvent): void {
...
-      token.owner = owner_address;
-
-      // Save both entities
-      owner.save();
-      token.save();
+    if (parseInt(event.params.to.toHexString()) === 0) {
+      // Burn
+      // Remove the entity from storage
+      // Its controllers and owner will be affected.
+      store.remove('Token', TokenId.toString());
```
jsonsivar, line 433: Do we also need to remove the current owner as an owner?

jsonsivar (approved):
LGTM. But once we merge #161, these changes would have to be done on `transfer.ts`.

subgraph/src/fleek-nfa.ts
EmperorOrokuSaki: We only need to save entities when their metadata have been modified. In this handler, owner is saved after the changes.

subgraph/src/fleek-nfa.ts
EmperorOrokuSaki: The owner entity can hold multiple access points, tokens, and also be an owner of the collection. When the token is removed, it will also affect the list of the tokens the owner holds since it uses a reverse lookup from the `owner` field of the Token entity.

jsonsivar (approved):

PR #166, opened 2023-03-08
feat: UI fetch mint price on contract

zoruka:


subgraph/src/fleek-nfa.ts
```diff
@@ -25,7 +25,7 @@ import {
   NewAccessPoint as NewAccessPointEvent,
   ChangeAccessPointNameVerify as ChangeAccessPointNameVerifyEvent,
   ChangeAccessPointContentVerify as ChangeAccessPointContentVerifyEvent,
-  TokenRolesCleared as TokenRolesClearedEvent
+  TokenRolesCleared as TokenRolesClearedEvent,
```
csosamorales, line 28: check this, I have the same problem on https://github.com/fleekxyz/non-fungible-apps/pull/173

csosamorales (approved):

ui/src/integrations/ethereum/hooks/ethereum-hooks.tsx
```diff
@@ -147,6 +156,8 @@ export namespace EthereumHooks {
       children?: React.ReactNode | React.ReactNode[];
       config?: ProviderConfig<TFunctionName>;
     }
+
+    export type SettingsParam = { value?: string };
```
jsonsivar, line 160: Does this ned also the other params like gas price etc.?

jsonsivar (approved):
LGTM (but I skipped subgraph changes since its just format)

ui/src/integrations/ethereum/hooks/ethereum-hooks.tsx
zoruka: We might add it as they are needed. I searched for a type aggregator from ethers library with everything that may go with this settings object, but as I could not find I'have added just what we are using.

PR #175, opened 2023-03-10
chore: UI clean up console logs

csosamorales:


ui/src/components/toast/toast.styles.tsx
```diff
@@ -85,6 +85,7 @@ export abstract class ToastStyles {
     display: 'flex',
     flexDirection: 'column',
     gap: '$4',
+    zIndex: 4,
```
zoruka, line 88: We can standardize the z-index in the theme file, this would help keeping track of what comes on top of what in the app, through the settings. It could be variables like `$toast`, `$modal`, `$background` and so on.

ui/src/views/mint/nfa-step/form-step/fields/logo-field/logo-field.tsx
```diff
@@ -16,9 +17,9 @@ export const LogoField = () => {
         const fileBase64 = await fileToBase64(file);
         setAppLogo(fileBase64);
         setErrorMessage(null);
-        //TODO remove console.log
+
         // To send to the contract the logo needs to be a base64 string
-        console.log('Sending to contract:', fileBase64);
+        AppLog.info('Sending to contract:', fileBase64);
```
zoruka, line 22: I think this may be removed, wdyt?

zoruka (approved):

ui/src/views/mint/nfa-step/form-step/fields/logo-field/logo-field.tsx
csosamorales: sure!

PR #180, opened 2023-03-14
chore: polygon to eth main migration

jsonsivar:


contracts/README.md
```diff
@@ -140,7 +140,7 @@ $ yarn deploy:hardhat
 
 If the execution is successful, you will see the contract address on your screen.
 
-### **Polygon Mumbai Testnet**
+### **Ethereum Sepolia Testnet**
```
zoruka, line 143: Should we change entirely for Sepolia or we could keep both in docs as we already have run through mumbai?

zoruka (approved):

contracts/README.md
jsonsivar: Ah good point, I vote for keeping it as well. I can change that.

PR #184, opened 2023-03-16
feat: contracts add verified field and allow verifier

zoruka:


contracts/contracts/FleekERC721.sol
```diff
@@ -49,6 +49,7 @@ contract FleekERC721 is
     uint256 private _appIds;
     mapping(uint256 => Token) private _apps;
     mapping(uint256 => address) private _tokenVerifier;
+    mapping(uint256 => bool) private _tokenVerified;
```
EmperorOrokuSaki, line 52: A mapping like this takes up a lot of storage and is very expensive. I recommend adding a verified field to the App struct and doing lookups from there.

EmperorOrokuSaki (changes requested):

jsonsivar (approved):
LGTM

PR #191, opened 2023-03-23
feat: UI simple create ap view

csosamorales:
- NFA pre-selected

https://user-images.githubusercontent.com/52458215/227960426-47612f41-aa86-4a5d-b323-89634b2b4d29.mov

- No NFA pre-selected

https://user-images.githubusercontent.com/52458215/227960440-aa299593-47b5-4a1d-810d-0f69f0e948d4.mov

### DIFF
--- a/ui/src/views/access-point/create-ap.form-body.tsx
+++ b/ui/src/views/access-point/create-ap.form-body.tsx
@@ -11,11 +11,8 @@ import { CreateAccessPoint } from './create-ap.context';
 
 export const CreateAccessPointFormBody = () => {
   const { id } = useParams();
-  const { setNfa } = CreateAccessPoint.useContext();
-
   const { nextStep } = Stepper.useContext();
-
-  const { nfa, billing } = CreateAccessPoint.useContext();
+  const { nfa, setNfa, billing } = CreateAccessPoint.useContext();
   const { setArgs } = CreateAccessPoint.useTransactionContext();
 
   const {


zoruka (changes requested):
Overall we need to be more accurate and descriptive with naming things. Keeping a good standard for all makes easier for other developers understand the code.

It is really important to be aware when passing params through url, users may input anything they want so we need to always validate and have fallback treatments.

ui/src/views/access-point/create-ap.form-body.tsx
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }, [nfaError]);
+
+  useEffect(() => {
+    if (nfaData) {
+      if (nfaData.token && id) {
+        const { name } = nfaData.token;
+        setNfa({ value: id, label: name });
+      } else {
+        AppLog.errorToast("We couldn't find the NFA you are looking for");
```
zoruka, line 57: We need to something else than just print it here. When we have the designs is good to at least clean up the search param in the URL and allow the user to select a NFA or something like that for this flow.

ui/src/views/access-point/create-ap.form-body.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { getNFADocument } from '@/graphclient';
+import { CreateAccessPoint } from './create-ap.context';
+
+export const CreateAccessPointFormBody = () => {
+  const { id } = useParams();
+  const { setNfa } = CreateAccessPoint.useContext();
+
+  const { nextStep } = Stepper.useContext();
+
+  const { nfa, billing } = CreateAccessPoint.useContext();
```
zoruka, line 18:
```suggestion
  const { nextStep } = Stepper.useContext();
  const { nfa, setNfa, billing } = CreateAccessPoint.useContext();
```

PR #193, opened 2023-03-23
feat: design ens verification flow

zoruka:


zoruka:
[sc-714]

zoruka:
[sc-714]

zoruka:
[sc-714]

zoruka:
[sc-714]

contracts/contracts/util/FleekENS.sol
```diff
@@ -0,0 +1,32 @@
...
+     */
+    function requireENSOwner(string calldata name) internal view {
+        if (_ens.owner(namehash(bytes(name), 0)) != msg.sender) revert MustBeENSOwner();
+    }
+
+    /**
+     * @dev Processes the name and returns the ENS node hash.
+     */
+    function namehash(bytes calldata name, uint256 index) internal view returns (bytes32) {
+        for (uint256 i = index; i < name.length; i++) {
```
jsonsivar, line 25: :cry:  (but I understand there is no way around this)

contracts/test/foundry/FleekERC721/AccessControl.t.sol
```diff
@@ -214,10 +215,12 @@ contract Test_FleekERC721_AccessControl is Test_FleekERC721_Base, Test_FleekERC7
...
 
         // TokenOwner
+        transferENS(ens, tokenOwner);
         vm.prank(tokenOwner);
         CuT.setTokenENS(tokenId, ens);
 
         // TokenController
+        transferENS(ens, tokenController);
         vm.prank(tokenController);
         CuT.setTokenENS(tokenId, ens);
```
jsonsivar, line 225: Should we not allow changing by controller? This can just be set by owner right?

jsonsivar (commented):
Just to recap our convo, did we say there was an issue with using subdomain to verify instead?

I.e., in order for someone to add the ENS, they first go to ENS and add a TXT record saying nfa.<rest of domain they want to add> (e.g., nfa.fleek.eth) and then we check for that. And that way we don't have to worry about transfers because it's up to the new recipient of the  ENS if they want to keep it or not.

zoruka:
> Just to recap our convo, did we say there was an issue with using subdomain to verify instead?
> 
> I.e., in order for someone to add the ENS, they first go to ENS and add a TXT record saying nfa. (e.g., nfa.fleek.eth) and then we check for that. And that way we don't have to worry about transfers because it's up to the new recipient of the ENS if they want to keep it or not.

I think I did not get exactly about the transfer part. This is a single time validation, by now we do nothing if the NFA or ENS is transferred. By the way it checks by owner of the domain/subdomain, if an ENS owner want someone to create e NFA using his domain, he will need to create a subdomain for this someone to use it (e.g. nfa.fleek.eth). I think this is the simples to way to verify because the owner doesn't need to change nothing in his ENS to be able to mint a NFA using it, what do you think?

contracts/test/foundry/FleekERC721/AccessControl.t.sol
zoruka: The way it is right now it is equals to the other properties that controllers may change. In my opinion is not bad that way because if someone else than the owner of the NFA is owner of the ENS, the owner of the NFA can add the owner of the ENS as a controller and he can call `setTokenENS` with his account. The other situation would be the owner of the ENS create a subdomain for the owner of the NFA, both ways works.

jsonsivar: Alright that's fair. I like the idea of the owner being able to add the owner of the ENS as a controller :+1:

jsonsivar:
> > Just to recap our convo, did we say there was an issue with using subdomain to verify instead?
> > I.e., in order for someone to add the ENS, they first go to ENS and add a TXT record saying nfa. (e.g., nfa.fleek.eth) and then we check for that. And that way we don't have to worry about transfers because it's up to the new recipient of the ENS if they want to keep it or not.
> 
> I think I did not get exactly about the transfer part. This is a single time validation, by now we do nothing if the NFA or ENS is transferred. By the way it checks by owner of the domain/subdomain, if an ENS owner want someone to create e NFA using his domain, he will need to create a subdomain for this someone to use it (e.g. nfa.fleek.eth). I think this is the simples to way to verify because the owner doesn't need to change nothing in his ENS to be able to mint a NFA using it, what do you think?

Yes, this definitely works for the mint part. I was wondering if the ENS gets transferred after the minting, then it might not map to the site anymore but the NFA will still point to the ENS. So wondering if we can address that somehow ...

jsonsivar (approved):
LGTM. Just closing the loop on this based on our DMs. Let's merge this and I made this card to track any extra steps we can take https://app.shortcut.com/fleekxyz/story/754/contracts-check-ens-on-queries-too.

contracts/contracts/util/FleekENS.sol
EmperorOrokuSaki: Minor gas optimization tip here: it's better for us to store `name.length` as a memory variable before the loop starts and use that var in the loop.
I had read about this before but I'm not fully sure if it's going to have a big impact here. Worth checking.

EmperorOrokuSaki:
> 

I think handling the transfer is a big piece. We can also leverage our substreams module to handle transfers. Please, let me know when this is being discussed next time so that I can join the discussion :)

EmperorOrokuSaki (approved):

PR #206, opened 2023-04-04
feat: CI/CD xyz

csosamorales:
Deploy to xyz [sc-757]


.github/workflows/fleek-deploy.yaml
```diff
@@ -0,0 +1,42 @@
...
+    paths:
+      - 'ui/**'
+
+jobs:
+  deploy-to-fleek:
+    runs-on: ubuntu-latest
+
+    env:
+      FLEEK_TOKEN: ${{ secrets.FLEEK_TOKEN }}
+      FLEEK_PROJECT_ID: ${{ secrets.FLEEK_PROJECT_ID }}
```
jsonsivar, line 17: Don't we need more env vars here?

.github/workflows/fleek-deploy.yaml
```diff
@@ -0,0 +1,42 @@
+name: Deploy site via Fleek
+
+on:
+  pull_request:
+    branches:
+      - main
```
jsonsivar, line 6: Should we do a separate deployment for main? I'm assuming the site ID in fleek.json is for the dev site.

csosamorales: ah, yeah! will separate it

ui/src/integrations/ethereum/ethereum.ts
```diff
@@ -32,6 +32,7 @@ export const Ethereum: Ethereum.Core = {
     return new ethers.Contract(contract.address, contract.abi, provider);
   },
 
+  //TODO remove cause we're using ENS subgraph
```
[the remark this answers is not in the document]
zoruka: we were using alchemy to fetch ENS names but we replaced for graphql, maybe is just to not forget to remove

PR #218, opened 2023-04-12
style: fix eslint errors for subgraph

EmperorOrokuSaki:


zoruka (approved):
LGTM. It has warnings on `subgraph/examples/query` but I think the project is outdated, should we create a card to update it in future?

EmperorOrokuSaki:
> LGTM. It has warnings on `subgraph/examples/query` but I think the project is outdated, should we create a card to update it in future?

I've fixed all warnings. The ones that are still there have reasons and some shouldn't really change :)

PR #219, opened 2023-04-12
feat: add log decoder helper and import web3 [serverless]

EmperorOrokuSaki:


serverless/src/libs/log-decoder.ts
```diff
@@ -0,0 +1,8 @@
+var Web3 = require('web3');
+
+var web3 = new Web3(Web3.givenProvider || "ws://localhost:17895");
```
zoruka, line 3:
```suggestion
const web3 = new Web3(Web3.givenProvider || "ws://localhost:17895");
```

serverless/src/libs/log-decoder.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export const logDecoder = (eventFieldsABI: { indexed: boolean; internalType: string; name: string; type: string; }[], data: string, topics: string[]) => {
+    return web3.eth.abi.decodeLog(eventFieldsABI, data, topics);
+};
```
zoruka, line 7:
```suggestion
export const logDecoder = web3.eth.abi.decodeLog;
```

jsonsivar (changes requested):
As discussed in scrum, let's close this and add web3 export to our solidity helper and we can use the web3 decoder function directly from that exported web3 instance.

EmperorOrokuSaki:
will be covered in #232

PR #222, opened 2023-04-13
fix: add alchemy keys missinng

csosamorales:


ui/src/components/core/input/input-file.styles.ts
```diff
@@ -16,7 +16,7 @@ export abstract class InputFileStyles {
     transition: 'border-color 0.2s ease-in-out',
     borderWidth: '$default',
     borderRadius: '$lg',
-    zIndex: '$docked',
+    zIndex: '$base',
```
zoruka, line 19: Does this element needs z-index? It seems to be static, z-index would be not applied.

ui/src/store/features/github/async-thunk/login.ts
```diff
@@ -45,6 +45,7 @@ export const login = createAsyncThunk(
         throw Error('Invalid response type');
       }
     } catch (error) {
+      AppLog.error(error);
```
zoruka, line 48:
```suggestion
      AppLog.errorToast('Github login failed. Please try again later.', error);
```
and remove the next line

zoruka (approved):

PR #224, opened 2023-04-14
feat: MongoDB & Prisma [serverless]

EmperorOrokuSaki:


serverless/prisma/tsconfig.json
zoruka, line 1: I think we might add this file in the root `.eslintrc.js` file as a project to recognize correctly the prisma project

zoruka, line 1: Do we need to keep all the comments here?

serverless/prisma/client/index.ts
```diff
@@ -0,0 +1,32 @@
...
+
+main()
+  .then(async () => {
+    await prisma.$disconnect()
+  })
+  .catch(async (e) => {
+    console.error(e)
+    await prisma.$disconnect()
+    process.exit(1)
+  })
```
zoruka, line 32:
```suggestion
main()
  .catch(async (e) => {
    console.error(e)
  })
  .finally(async () => {
    await prisma.$disconnect();
  });
```

serverless/prisma/client/index.ts
```diff
@@ -0,0 +1,32 @@
+import { PrismaClient } from '@prisma/client'
+
+const prisma = new PrismaClient()
+
+async function main() {
```
jsonsivar, line 5: I don't know if this lib needs a "main" function.  We should create a function called `getAllTokens` or similar and put this logic there.

jsonsivar (dismissed):
Hey for this I think we should make prisma a lib in the serverless package and not create another sub-package? Ie I dont think prisma folder needs its own package.json or tsconfig.json.

serverless/prisma/client/index.ts
zoruka: maybe just leave a comment saying that is a example is enough

zoruka (approved):

PR #232, opened 2023-04-19
feat: create smart contract helper for serverless

EmperorOrokuSaki:


serverless/src/functions/builds/handler.ts
```diff
@@ -11,6 +12,14 @@ export const submitBuildInfo = async (): Promise<APIGatewayProxyResult> => {
       createdAt: new Date().toISOString(),
     };
 
+    nfaContract.methods
+      .setTokenBuild(1, 'hash', 'repo')
+      .call((err: string | undefined, res: any) => {
+        if (err) throw new Error(err);
+        console.log('result');
+        console.log(res);
+      });
```
zoruka, line 21: I wonder how the gas is going to be paid and also it needs to be signed.

zoruka (approved):
Maybe it is a good idea leave comments saying where it is examples

serverless/src/functions/builds/handler.ts
EmperorOrokuSaki: this is a placeholder.

PR #233, opened 2023-04-19
test: remove all matchstick tests

EmperorOrokuSaki:


jsonsivar (approved):
LGTM.  Let's make sure we add a card in shortcut to track how we test these then.

PR #235, opened 2023-04-20
feat: subgraph updated handlers

zoruka:


zoruka:
[sc-832]

subgraph/src/constants.ts
```diff
@@ -0,0 +1,10 @@
+export const CollectionId = 'root';
```
EmperorOrokuSaki, line 1: Is this supposed to be the contract address?

subgraph/src/mint.ts
```diff
@@ -68,13 +75,37 @@ export function handleNewMint(event: NewMintEvent): void {
...
+  const collection = Collection.load(CollectionId);
+  if (collection) {
+    collection.totalTokens = collection.totalTokens.plus(BigInt.fromU32(1));
+    collection.save();
+  }
 
   // Save entities
   owner.save();
   token.save();
+  repository.save();
```
EmperorOrokuSaki, line 110: verifier should be saved too.

EmperorOrokuSaki (changes requested):

subgraph/src/constants.ts
zoruka: No, this is just a keyword to make simple for requests, we may use whatever we want. I just think using variable thing like contract address, collection name or whatever would just overcomplicate stuff for both reading/updating this unique value in the handlers and for the client having to input a variable value to read this data.

subgraph/src/mint.ts
EmperorOrokuSaki: Update on this from my discord chat with @zoruka :

Since the verifier is not created here, checking its existence will be enough.

subgraph/src/constants.ts
EmperorOrokuSaki: I understand your point here. However, I personally don't like using root because it requires external knowledge of the subgraph to work with it. Other services or devs who might interact with the subgraph in the future won't know what keyword they should use.

The contract address is what I had initially used in the `Collection` entity that was later removed. But I'm open to other ideas.

zoruka: I think we should define a keyword that's easy to be a constant for both sides. Even if we use the contract id it still needs external knowledge to know that the id is going to be the contract id. We can just have this described in docs. Besides that, what you suggest?

EmperorOrokuSaki (approved):

PR #237, opened 2023-04-24
docs: update readme, add serverless section.

EmperorOrokuSaki:


README.md
```diff
@@ -1,6 +1,6 @@
-# Fleek Non Fungible Apps
+# Fleek Non-Fungible Apps
```
zoruka, line 1: I think adding `-` should be double checked with marketing side of the project before merging it

zoruka (approved):

README.md
EmperorOrokuSaki: I confirmed the hyphened version with Nico.

PR #245, opened 2023-04-26
feat: UI responsiveness on mint and AP flow

csosamorales:


ui/src/views/mint/github-step/steps/github-connect/github-connect-step.tsx
```diff
@@ -7,13 +7,18 @@ export const GithubConnect: React.FC = () => {
   const { prevStep } = Stepper.useContext();
 
   return (
-    <Card.Container>
+    <Card.Container css={{ maxWidth: '$107h' }}>
```
zoruka, line 10: Seems like all `Card.Container` usages uses this same `maxWidth`, should it be a property from the component?

ui/src/views/mint/github-step/steps/github-repo-configuration/repo-configuration-body/repo-branch-commit-fields.styles.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const TextStyles = styled(Text, {
+  maxWidth: '12.5rem',
+
+  '@media (min-width: 375px)': {
+    maxWidth: '17rem',
+  },
+  '@media (min-width: 425px)': {
+    maxWidth: '18rem',
+  },
```
zoruka, line 12: I think here it should fulfill the parent, it should not? maybe using `100%` would work better than setting fixed values.

zoruka (approved):

ui/src/views/mint/github-step/steps/github-connect/github-connect-step.tsx
csosamorales: Yes, I was thinking to work on that refactor today (Card Header and this Card component that is for mint and access point flows)

PR #262, opened 2023-05-17
feat: lambda layers

EmperorOrokuSaki:


contracts/hardhat.config.ts
```diff
@@ -86,10 +87,10 @@ const config: HardhatUserConfig = {
...
   },
   etherscan: {
-    // apiKey: {
-    //   polygonMumbai: POLYGONSCAN_KEY,
-    // },
-    apiKey: ETHERSCAN_API_KEY ? ETHERSCAN_API_KEY : '',
+    apiKey: {
+      polygonMumbai: POLYGONSCAN_KEY,
+    },
+    // apiKey: ETHERSCAN_API_KEY ? ETHERSCAN_API_KEY : '',
```
zoruka, line 93: maybe we are able to add for both networks?

zoruka (approved):
About js side everything looks good to me, but I'm not deep sure about the layers settings. Maybe @jsonsivar can help with that.

contracts/hardhat.config.ts
EmperorOrokuSaki: done

PR #264, opened 2023-05-19
feat: bunny cdn integration and create pull zone

EmperorOrokuSaki:


serverless/src/functions/apps/handler.ts
```diff
@@ -0,0 +1,106 @@
...
+        if (
+          error instanceof BunnyCdnError &&
+          error.name === 'pullzone.name_taken'
+        ) {
+          continue;
+        } else {
+          throw error;
+        }
+      }
+    } while (callSuccess === false);
```
zoruka, line 65: I understand that this is not expected to, but may result into infinite loop. Maybe would be a good idea create max number of tries exception or something like that.

serverless/src/libs/bunnyCDN.ts
```diff
@@ -0,0 +1,261 @@
...
+  ({ name }: FetchPullZoneWhenNameAlreadyTakenArgs) =>
+  async (error: BunnyCdnError | AxiosError<ErrorData>) => {
+    if (
+      error instanceof BunnyCdnError &&
+      error.name === 'pullzone.name_taken'
+    ) {
+      return fetchPullZone({ name });
+    }
+
+    throw error;
```
zoruka, line 246: this code seems to be repeated with the previous comment, a possible enhancement would be a single function that repeats for this given error.

zoruka (approved):

serverless/src/functions/apps/handler.ts
EmperorOrokuSaki: good point

serverless/src/libs/bunnyCDN.ts
EmperorOrokuSaki: this function returns the pullzone but in the previous comment's code snippet we continue to the next loop iteration.
Actually, this function should be removed since we don't use it anywhere.

serverless/src/functions/apps/handler.ts
```diff
@@ -0,0 +1,107 @@
...
+  try {
+    // Check the parameters and environment variables
+    dotenv.config();
+    if (event.body === null || process.env.BUNNY_CDN_ACCESS_KEY == undefined) {
+      return formatJSONResponse({
+        status: 422,
+        message: 'Required parameters were not passed.',
+      });
+    }
+
```
jsonsivar, line 24: Should we check if the app existing in the DB before we create a new one? Or will the create at the bottom of the file do an upsert?

EmperorOrokuSaki: Line 78 should check that.

PR #266, opened 2023-05-22
feat: AWS Lambda deployment script and fixes

EmperorOrokuSaki:


serverless/src/functions/builds/handler.ts
```diff
@@ -27,7 +27,8 @@ export const submitBuildInfo = async (
       ipfsHash: data.ipfsHash,
       domain: data.domain,
     };
-
+    console.log(buildInfo);
```
zoruka, line 30: maybe leftover

serverless/src/functions/builds/handler.ts
```diff
@@ -37,8 +38,12 @@ export const submitBuildInfo = async (
         domain: buildInfo.domain,
       },
     });
+    console.log(buildRecord);
```
zoruka, line 41: maybe leftover

serverless/src/functions/builds/handler.ts
```diff
@@ -37,8 +38,12 @@ export const submitBuildInfo = async (
...
+    
 
     if (buildRecord.length == 0) {
+      console.log('here i am');
```
zoruka, line 45: maybe leftover

zoruka (approved):
It is good review if all the commented envs should stay like that. Also I would request @jsonsivar to eyeball this one.

serverless/src/functions/builds/handler.ts
EmperorOrokuSaki: I'm not quite sure if we want to keep this or not. But I guess we can add it whenever needed, later.

serverless/src/functions/builds/handler.ts
EmperorOrokuSaki: These should be overwritten by the #272

serverless/src/functions/builds/handler.ts
EmperorOrokuSaki: These should be overwritten by #272

PR #271, opened 2023-06-02
feat: verify hostname (load free certificate call to bunny cdn)

EmperorOrokuSaki:


serverless/src/functions/apps/handler.ts
```diff
@@ -0,0 +1,40 @@
...
+    const bunnyCdn = new BunnyCdn(process.env.BUNNY_CDN_ACCESS_KEY);
+    const hostname = JSON.parse(event.body).hostname;
+
+    let args: LoadFreeCertificateMethodArgs = {
+      hostname,
+    };
+
+    await bunnyCdn.loadFreeCertificate(args).catch((e) => {
+      throw e;
+    });
```
zoruka, line 29: This code is the same than:
```suggestion
    await bunnyCdn.loadFreeCertificate(args);
```

serverless/src/libs/bunnyCDN.ts
```diff
@@ -0,0 +1,261 @@
...
+  public name: string;
+
+  constructor({ name, message }: BunnyCdnErrorOptions) {
+    super(message);
+
+    this.name = name;
+    this.message = message;
+  }
+}
+const BUNNY_CDN_API_URL = 'https://api.bunny.net';
```
zoruka, line 18: should this be from `env`?

zoruka (approved):
this PR comes with changes related to #264, maybe check if need to merge both

serverless/src/libs/bunnyCDN.ts
EmperorOrokuSaki: It's not going to change.

PR #278, opened 2023-06-16
feat: dev environment backend setup

EmperorOrokuSaki:


zoruka (approved):

contracts/deployments/qanet/FleekERC721.json
zoruka, line 1: do we need the deployment files from qanet to be pushed with the repo code?

serverless/src/functions/apps/handler.ts
```diff
@@ -75,7 +71,11 @@ export const submitAppInfo = async (
   try {
     // Check the parameters and environment variables
     dotenv.config();
-    if (event.body === null || process.env.BUNNY_CDN_ACCESS_KEY == undefined || event.headers.originUrl === undefined) {
+    if (
+      event.body === null ||
+      process.env.BUNNY_CDN_ACCESS_KEY == undefined ||
```
zoruka, line 76:
```suggestion
      process.env.BUNNY_CDN_ACCESS_KEY === undefined ||
```

zoruka (approved):

contracts/deployments/qanet/FleekERC721.json
EmperorOrokuSaki: I don't have a strong opinion on this. Should they not be pushed?

serverless/src/functions/apps/handler.ts
EmperorOrokuSaki: implemented

PR #280, opened 2023-06-29
docs: main readme corrections

nicolaspoggi:


README.md
```diff
@@ -1,20 +1,22 @@
...
 
-**The repository for Fleek Non-Fungible Apps project**
+# Non-Fungible Apps - by Fleek
 
 > 🚧 IMPORTANT 🚧 - This initiative is under development, so this repo should be treated as a WIP. The goals and the roadmap might change as the project is shaped.
 
 ## ⚡ Overview
 
-This is the landing point for Fleek's initiative to implement infrastructure as Solidity contracts.
+This is the landing point for our initiative to decentralize web3 app frontend infrastructure, their distribution, and access, through Non-fungible Applications, an NFT-based implementation.
```
EmperorOrokuSaki, line 8: The current implementation supports frontends, but the long-term goal is to add support for everything an app has (as much as we can at least, like multi-sigs and such)
Should that be included here?

README.md
```diff
@@ -1,20 +1,22 @@
...
 
-The vision is to have this on Ethereum Mainnet. We can create a network of smart contracts that represent the different parts of your stack. We had previously targeted Polygon but have changed it to Ethereum Mainnet for better interoperability and the security that mainnet provides.
+**The goal of NFAs** is to put web3 application frontends on-chain to provide users with a verifiable and crypto-friendly way of accessing applications.
```
EmperorOrokuSaki, line 10: My question in the comment above applies to this line as well.

README.md
```diff
@@ -1,20 +1,22 @@
...
 
-The goal is to be a more verifiable and crypto-friendly Serverless.yaml or Cloudformation manifests that will enable us to develop use cases on top like community hosting.
+1. Developers mint their app's frontend onto an NFT, where the frontend bundle is stored and referenced from IPFS.
+2. Users mint prints of the NFA, which allows them to have a decentralized and permanently on-chain access point to any web3 app.
```
EmperorOrokuSaki, line 13: We have decided to use the term `apps` instead of `prints`

EmperorOrokuSaki (changes requested):
Looks good! Added some comments and left some questions in the review 😄

README.md
nicolaspoggi: the word infrastructure here I think covers it, but in the one below I will do edit

README.md
nicolaspoggi: Let's bring to broader convo, I think we are using prints elsewhere atm so for consistency I leave it - to not confuse with "NFA print, NFA app, Nonfungible application apps" can sound redundant initially.

EmperorOrokuSaki (approved):

EmperorOrokuSaki:
Merged as #281

